PR #14, opened 2015-10-14
Three new resources:

raviii:
1. Customer Address(with a test case)
2. Policy(with a test case)
3. Carrier Service(No test case)

-Locally tested CustomerAddress and Policy resources from Nodify-App.
-Could not test CarrierService as its only enabled for Plus Customers.


Alua-Kinzhebayeva:
Hi @raviii!
I have pulled and tested your branch new_resources.
I have also pushed some fixes to https://github.com/Alua-Kinzhebayeva/Nodify/commits/new_resources so you could see what I am talking about.

**Customer Adress resource**:
While what you did technically works with  CustomerAddress.prototype.slug = "addresse";  and current implementation of endpoints of the resource (it might break endpoints with "address" as return object), it would have been stylistically better if we left slug as "address" and overwrite "all" as   
```
CustomerAddress.prototype.all = function(parentId, params, callback) {
      var url, _ref;
      if (typeof params === 'function') {
        _ref = [callback, params], params = _ref[0], callback = _ref[1];
      }
      url = this.resource.queryString("" + this.prefix + "/" + parentId + this.child, params);
      return this.resource.get(url, "" + this.slug, callback);
    };
```
**Carrier Service resource**:
I wrote integration test and changed slug to slug: "carrier_service" from slug: "carrier_services", because it would break its parent's methods (Base.js) implementation like "all" for example (it would have appended extra "s" => carrier_servicess)

raviii:
Hi @Alua-Kinzhebayeva ,

Thanks for that!

Your fix for carrier_service totally make sense.
Just started with coffee script and somehow didnt see that coming.

Although with address'e's bug, I was just debugging it till late and was looking to just make it work. I knew it was a bad idea!

Alua-Kinzhebayeva:
:relaxed:

Issue #26, opened 2015-10-28
complete User resource 


Alua-Kinzhebayeva:
complete [User](https://docs.shopify.com/api/user) with integration tests

PR #30, opened 2015-11-05
Add User resource

khanghoang:
Complete #26


khanghoang:
@Alua-Kinzhebayeva I will make it better in the future PRs, because this PR does 2 things instead of focus in one thing. And It should be on another branch which its name describes what It does instead of "master"

Alua-Kinzhebayeva:
@khanghoang alright! thank you for the pull request.

PR #36, opened 2015-11-16
session for private apps

Alua-Kinzhebayeva:
added support for creating a session for private apps


alexandresaiz:
let's go!!!!!! I eventually can have the pleasure to merge a PR!!

PR #53, opened 2015-11-18
Add LICENSE file

lpinca:
This adds a proper `LICENSE` file that will be included in the npm package.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,22 @@
+(The MIT License)
+
+Copyright (c) 2012-2015 microapps (http://www.microapps.com/?utm_source=nodify-module-repo-readme&utm_medium=click&utm_campaign=github)
+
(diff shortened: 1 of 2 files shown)


alexandresaiz:
@lpinca not a big deal, but we normally want to have ANY microapps url tagged. When viewing this readme file from a browser within GH it's indeed clickable. Anyway, Analytics does track the referer too. It's just a matter of following the same concept always.

lpinca:
@alexandresaiz This is how you'll see it on GitHub https://github.com/lpinca/Nodify/blob/ce413f15cdabfd4c749aa64a2ab80d131ecc3191/LICENSE

lpinca:
Anyway no problem, I will add the query string back.

lpinca:
Done.

alexandresaiz:
@lpinca I was confused between the readme and the license file. Sorry.

PR #60, opened 2015-11-19
Fix badges

lpinca:
This fixes the markdown for the badges.


lpinca:
Ok done with rebasing, I hope we don't step on each other's feet anymore :smile:

PR #61, opened 2015-11-19
Update contributors

lpinca:
This updates the list of contributors.


lpinca:
I messed up the format, better to use a bullet list, will fix in a bit.

alexandresaiz:
:+1:

PR #62, opened 2015-11-19
Add pre-commit to automatically run the tests before every commit

lpinca:
This adds [`pre-commit`](https://github.com/observing/pre-commit).
It will automatically run the tests before every commit and it will prevent changes from being commited if tests fail.

It will also be useful in future to run the lint checks before every commit.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "mocha": "2.3.x",
     "nock": "3.0.x",
     "paige": "0.1.x",
+    "pre-commit": "1.1.x",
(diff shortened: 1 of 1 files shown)


Alua-Kinzhebayeva:
does it work without "pre-commit": {
    "run": ["foo", "bar", "test"],
  } ?

lpinca:
Yes, by default it runs the `test` script

Alua-Kinzhebayeva:
fine then

PR #67, opened 2015-11-28
Recompile the coffee files and downgrade qs to version 5.2.x

lpinca:
This fixes the build.


lpinca:
Almost :smile:

lpinca:
Ok, fixed now, `qs@6` only works on [`node >= 4`](https://github.com/hapijs/qs/blob/master/package.json#L16).

PR #75, opened 2016-01-18
BUG: RecurringApplicationCharges.Activate returns an undefined body

markpete:
The following code throws in resource.js because the response from Shopify is a null body but a 200 OK response.

export function acceptChargeCallback(id, chargeRecord, session, res) {
    session.recurringApplicationCharge.activate(id, {
        "recurring_application_charge": chargeRecord
    }, function (result) {
        res.redirect("/");
    });
}

### DIFF
--- a/lib/resource.js
+++ b/lib/resource.js
@@ -50,8 +50,8 @@
         }
         if (method === 'DELETE') {
           body = slug;
-        } else if (slug && slug !== 'oauth') {
-          body = body[typeof slug === 'object' ? slug.long : slug];
+        } else if (slug && slug !== 'oauth' && typeof(body) === "object") {
+            body = body[typeof slug === 'object' ? slug.long : slug];
         }
         return process.nextTick(function() {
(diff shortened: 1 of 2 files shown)


lpinca:
LGTM but the change should be made in `resource.coffee`.

lpinca:
@markpete if you will update it, please squash the commits. Thank you.

markpete:
@lpinca  Change made to resource.coffee and commits squashed.  Please let me know if there are other changes you need me to make.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "nodify-shopify",
-  "version": "0.2.1",
+  "version": "0.2.2",
```
lpinca: @markpete sorry to pester, please remove this change.

markpete:
@lpinca  No bother at all.  I reverted the library version bump.  Basically all that's left is the null check.

lpinca:
Thank you!

PR #78, opened 2016-02-11
Move everything to ES6, small fixes, builds, test coverage, jsdoc

ivaylopivanov:
## Disclaimer
All of the code changes and implementation are made based on assumptions and guessing as I don't have first hand experience with the Shopify API

## Changes
1. All of the existing code is moved from coffee to ES6
2. All of the resources are returning `bluebird` promises
3. All of the existing tests are changed to consume a promise instead of a callback
4. All of the resources are extending the base classes as there is sense for that. The base classes are accessing statically an abstract adapter class to communicate with the Shopify API.
5. Small changes has been made on the session implementation, but I need feedback and examples in order to refactor it completely. I don't want to create more breaking changes before I get any feedback and to see how you're using it in production, and why you've chosen the existing control flow.
6. A lot of ambiguous, incorrect, and imperative code has been removed. A lot of it is still there unfortunately. Just a few examples:
 * nock returning status code 201 when method is delete
 * return statement after throw
 * params != null
 * method for sorting an object returning undefined: undefined
 * etc.
7. Updated package.json with node version for engines
8. Grunt build (including ESLint output for `./lib` dir) and tests automation:
```sh
$ npm install
$ npm run build
```
On file change, grunt will run eslint and tests so you will be able to see the output in the console
9. pre-commit for tests automation
10. Comment templates for generating documentation using jsdoc -
```sh
$ jsdoc ./lib
```
The output will be in the `out/` directory.
I didn't put any description about the resources as I don't want to put incorrect information about them - experience needed
11. Coverage and coveralls scripts in package.json for tests coverage
```sh
$ npm run coverage
```
The output will be in the `coverage/` directory.

## What is left
1. Refactoring of `./lib/sessions`
2. Description for all resources
3. Up to date readme, pointing to the existing doc.

## Suggestions
1. Create an account and add a badge for tests coverage - I prefer `coveralls.io`
2. I saw that a `gh-pages` branch exists, so a script can be put to generate the docs and checkout to that branch automatically
3. Verify the persistent_token (if there is such) when the session is created as it's ambiguous to receive "session is valid" while creating a session instance and then a status code `401` when accessing a resource.

## Summary
This pull request may or may not lead to more bugs and the implementation may or may not be the best possible, but will make it possible to maintain it without coffeescript. Fortunately, more node.js developers will step in and will try to contribute when the package is written in pure javascript.


lpinca:
Wow, awesome. It's hard to review such a big pr.

.eslintrc
```diff
@@ -0,0 +1,130 @@
+{
```
lpinca: People at microapps use this config:
```json
{
  "env": {
    "mocha": true
  },
  "rules": {
    "comma-dangle": [2, "never"],
    "object-curly-spacing": 0,
    "no-shadow": 0
  },
  "extends":"airbnb/base"
}
```
but we adjust this at a later stage it's not very important right now.

Gruntfile.js
```diff
@@ -0,0 +1,34 @@
+module.exports = function(grunt) {
```
lpinca: I would like to not use Grunt or Gulp, only npm scripts.

lib/sessions/session-oauth.js
```diff
@@ -0,0 +1,162 @@
+/**
```
lpinca: I would like to remove this completely. I think it's better to have this as a separate module, see https://github.com/lpinca/shopify-token.

lib/sessions/session.js
```diff
@@ -0,0 +1,146 @@
+/**
```
lpinca: See https://github.com/microapps/Nodify-Shopify/pull/78/files#r52592014.

package.json
```diff
@@ -7,29 +7,39 @@
...
+    "bluebird": "^3.2.1",
     "pluralize": "1.2.x",
     "qs": "5.2.x",
     "request": "2.9.x",
     "singleton": "1.0.x"
   },
   "devDependencies": {
-    "coffee-script": "1.10.x",
-    "docco": "0.7.x",
+    "coveralls": "^2.11.6",
```
lpinca: I prefer to install and run this in travis `after_script` hook. See https://github.com/lpinca/shopify-token/blob/b2ce543e1f84de9cfc0acc6d661ead4f11107409/.travis.yml#L11-L12.

lib/components/adapter.js
```diff
@@ -0,0 +1,124 @@
+/**
+ * @private
+ * @class Adapter
+ * @description Making requests to Shopify API.
+ * @requires request
+ * @requires bluebird
+ */
+'use strict';
+const request = require('request');
+const bluebird = require('bluebird');
```
lpinca: If I'm not mistaken `bluebird` is only used to promisify `request`, so we can probably use [`got`](https://github.com/sindresorhus/got) instead which has built in support for promises and is less bloated.

.eslintrc
ivaylopivanov: personal preferences - you can adjust them by your taste later

Gruntfile.js
ivaylopivanov: personal preferences - you can remove it completely if you want

lib/sessions/session-oauth.js
ivaylopivanov: I agree, as I stated earlier the whole session/ dir is a mess so you can remove it or implemented as you prefer

package.json
ivaylopivanov: I've missed those and I think it's appropriate to put coveralls into package.json instead of travis config file because it's nature to put the dependencies there but once again - personal preferences - you can adjust it as you want : )

lib/components/adapter.js
ivaylopivanov: Yes, it's only there. I've been wondering between native promises or bluebird, but I chose bluebird because of it's performance, since this is a hot sport in the application.

lpinca: The overhead of using native promises is minimal in this case, it's not something to worry about IMO.

lpinca:
@ivaylopivanov I've pushed a new branch called `rewrite`. Can you please resubmit there? Thank you.

ivaylopivanov:
Ok

PR #80, opened 2016-02-15
added checkout Resource

PauloLuan:


test/checkout_test.js
```diff
@@ -1,41 +1,52 @@
...
+  it('should get all checkouts', function (done) {
+    var resBody = fixtures.allResponseBody;
+
+    scope.get('/admin/checkouts.json')
+      .reply(200, resBody);
+
+    resource.all(function (err, res) {
+      if (err) return done(err);
+
       res.should.not.be.empty;
```
lpinca: Replace this line and the next with `res.should.be.eql(resBody.checkouts)`.

PR #82, opened 2016-02-17
added Shipping Zone resource

kazinov:


test/shipping_zone_test.js
```diff
@@ -0,0 +1,28 @@
+var helper = require("./common.js");
+helper.setObject("shipping_zone");
+
+var Resource = helper.resource();
+
+helper.nock(helper.test_shop)
+  .get('/admin/shipping_zones.json')
+  .reply(200, helper.load("all"), {
```
lpinca: Save the fixture somewhere, so we can reuse it later and remove the third argument.

test/shipping_zone_test.js
```diff
@@ -0,0 +1,28 @@
...
+    status: '200 OK'
+  });
+
+describe('Shipping zone', function() {
+  var site = helper.endpoint;
+  var resource = new Resource(site);
+
+  it('should get all shipping zones', function(done) {
+    resource.all(function(err, res){
+      res.should.not.be.empty;
```
lpinca: Use the previously saved fixture, e.g.`res.should.be.eql(fixture.shipping_zones)`;

kazinov: @lpinca, fixed

lpinca: Please also remove `res.should.not.be.empty;`, replace it with `if (err) return done(err);`

kazinov:
@lpinca fixed

lpinca:
Awesome, thank you.

PR #83, opened 2016-02-17
Add linting

kazinov:
This adds basic [ESLint](http://eslint.org/) configuration. It could be the first step for having consistent code style. Once the rules are discussed and this PR merged it would be easy to add an npm script for linting and make code cleanup according the @lpinca [suggestion](https://github.com/microapps/Nodify-Shopify/issues/57).

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -4,6 +4,10 @@
         "commonjs": true,
         "node": true
     },
+    "globals": {
+        "it": true,
+        "describe": true
+    },
     "extends": "eslint:recommended",
     "rules": {
         "indent": [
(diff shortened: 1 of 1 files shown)


lpinca:
@kazinov, thank you.
There is a work in progress to move everything to ES2015, see the `no-token-retrieval` branch. We already have a basic ESLint [configuration](https://github.com/microapps/Nodify-Shopify/blob/78ae29266ba7aab2d87b7ce46fa808214ded880a/.eslintrc) and a lint script in that branch.

To make the merging process a little easier I think I'm going to pass on this.

kazinov:
@lpinca, ok, I will close it

PR #90, opened 2016-04-22
API call limit support

SEAPUNK:
This adds a property to the shopify class called `callLimits`, which starts like this:

```js
{
  max: undefined,
  current: undefined,
  remaining: undefined
}
```

and after each request, looks like this:

```js
{
  max: 40,
  current: 6,
  remaining: 34
}
```

If the `x-shopify-shop-api-call-limit` header does not exist (is not a string), values don't get updated.

Ref: https://docs.shopify.com/api/guides/api-call-limit#how-to-avoid-the-429-error

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,20 @@ shopify.order.list({ limit: 5 })
   .catch(err => console.error(err));
 ```
 
+You can access your current call limits with `shopify.callLimits`.
+This is updated with each request to Shopify's API. Values start at undefined.
+
+```js
+console.log(shopify.callLimits)
+/*
+{
+  max: 40,
+  current: 10,
+  remaining: 30
+}
+*/
+```
+
 ### Available resources and methods
 
 - applicationCharge
--- a/index.js
+++ b/index.js
@@ -32,6 +32,15 @@ function Shopify(shop, key, password) {
     this.token = key;
   }
 
+  //
+  // API call limits, updated with each request.
+  //
+  this.callLimits = {
+    max: undefined,
+    current: undefined,
+    remaining: undefined
+  };
+
   this.baseUrl = {
     hostname: `${shop}.myshopify.com`,
     protocol: 'https:',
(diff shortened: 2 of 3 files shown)


SEAPUNK:
Coverage decreased because I'm not sure how I could go about writing a test for this. What do you recommend?

lpinca:
Hi @SEAPUNK,
you can send the `x-shopify-shop-api-call-limit` with nock, see https://github.com/node-nock/nock#specifying-reply-headers.

Using that feature you can add a test in `shopify.test.js`.

README.md
```diff
@@ -60,6 +60,19 @@ shopify.order.list({ limit: 5 })
   .catch(err => console.error(err));
 ```
 
+You can access your current call limits with `shopify.callLimits`. This is updated with each request to Shopify's API. Values start at undefined.
```
lpinca: Respect the 80 chars limit if possible.

index.js
```diff
@@ -69,6 +78,16 @@ Shopify.prototype.request = function request(url, method, key, params) {
   return got(options).then(res => {
     const body = res.body;
 
+    const callLimitHeader = res.headers['x-shopify-shop-api-call-limit'];
+
+    if (typeof callLimitHeader === 'string') {
```
lpinca: This can only be `undefined` or a string right? In that case I would go with just `callLimitHeader`.

SEAPUNK: As far as I know, it should be only undefined or a string. The typeof check is just a safety check in case anything goes awry.

I'll change it in the next squashed commit.

README.md
SEAPUNK: Fixed.

test/shopify.test.js
```diff
@@ -215,4 +215,44 @@ describe('Shopify', () => {
         .then(res => expect(res).to.deep.equal({}));
     });
   });
+
+  describe('Shopify#callLimits', () => {
```
lpinca: Sorry for being picky but this is not a method, it's a feature in the request method.
Can you please move these test in the above describe?

Thank you!

SEAPUNK: It's fine, I'm picky too. :)

I moved the first callLimits test (for undefined values) to the very beginning (to assure that no requests are done before this check), and the rest to the bottom. I also improved the wording of those three tests.

lpinca:
Thank you!

Issue #92, opened 2016-05-12
Release new version


SEAPUNK:
Could a new version be created and published, please? A few changes have been made since then, and I'd like to start utilizing those features (the API call limits in specific) without having to install the package via Github repo.

alexandresaiz:
@SEAPUNK feel free to create the new version. This is an Open Source project maintained by contributors like you!

PR #93, opened 2016-05-12
[dist] 1.1.0

SEAPUNK:
Changelog, from what I am looking at in the commits:

* Add TaxService API support (lpinca) 
* Add UsageCharge API support (lpinca) 
* Add ability to check API call limit (SEAPUNK & lpinca)

Ref: #92

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "shopify-api-node",
-  "version": "1.0.0",
+  "version": "1.1.0",
   "description": "Shopify API bindings for Node.js",
   "main": "index.js",
   "directories": {


SEAPUNK:
Wait, shoot. Wrong version. Let me redo that really quickly.

lpinca:
Thanks but the release is on hold for a reason.
I've assumed that `shopify.taxService.delete()` returns an [empty object](https://github.com/microapps/Shopify-api-node/commit/8c4cef68ce9c79f6f76e59118550a125e879b6b9#diff-a26185424d7ff01fdff1cf2283e688d6R66) like all other resources.

According to the [documentation](https://help.shopify.com/api/reference/taxservice) this is not true and the full object is returned.

We tried to to test this on a PLUS shop without success as this resource is new and doesn't work yet.
We would like to have some clarity about this before releasing 1.1.0.

If you need the API call limit feature please install from GitHub

SEAPUNK:
Alright, I'll close this PR then, since you know when this library will be ready for 1.1.0.

To be honest, I didn't expect @alexandresaiz to [tell me to create the version](https://github.com/microapps/Shopify-api-node/issues/92#issuecomment-218858292) -- I was hoping that the maintainers would do it when they feel that a new version should be released, since I'm not really in charge of these sorts of things.

lpinca:
Yes, sorry about that.

alexandresaiz:
Well, I felt like some teacher coming here and asigning homework. Sorry.

PR #95, opened 2016-06-23
update readme

ayhansipahi:
npm install script fixed


lpinca:
It also works as is. Any particular reason for this change?

ayhansipahi:
I got momentary error.
retried works as it should be.

lpinca:
Ok, thank you.

PR #97, opened 2016-08-05
[major] Construct with options object, instead of argument list

SEAPUNK:
Precursor to adding the `timeout` parameter to options object.

Ref: #96

Issue #96, opened 2016-08-02
Request timeouts


SEAPUNK:
IIRC, `got` doesn't have a default timeout, which seems to mean that a request can take forever if the program missed or just has not received a response. One of my programs that I wrote that utilizes this module seems to be getting bit by this problem.

I do not see a place where I can configure a timeout in this module; will support for this be added?

lpinca:
Yes it makes sense. Ideally we should make this an option, fall back to a default value if the option is not provided and forward that value to got. I'm not totally sure what interface to use though.
We can add an additional options argument to the constructor or change the construct to only accept an options object (this is a breaking change).
Another option is to add a `configure` method to allow configuration once the instance has been created but meh I'm not really happy with this.

SEAPUNK:
Ultimately, I personally prefer the breaking change to be implemented, which would then deconstruct the object into instance properties, where `requestOptions` would be one of them, which would be an object that gets additionally applied into the generated request option object, as kind of an override.

```js
class API {
  constructor (options) {
    this.requestOptions = options
  }

  request () {
    let options = assign(
        // ...
    )
    if (this.requestOptions) {
      options = assign(options, this.requestOptions)
    }
    // [...]
  }
}
```

...but because you probably don't want to immediately make the breaking change, a non-breaking change that would be a step towards such a change would be just doing the instance property of `this.requestOptions`, document it, and have people set it immediately after construction. It's not pretty, but it's also possibly not going to be used often, and to me is the cleanest way to do this if we're looking at the big picture.

```js
const api = new API()
api.requestOptions = {timeout: 30000}
```

Also, would setting a default request timeout be considered a breaking change, feature addition, or a bug fix?

lpinca:
It's a fix and yes the breaking change is probably the cleanest. :-1: on using properties directly added by the user like `shopify.requestOptions` it doens't matter if it's a setter.
The module has only ~500 downloads per month and only 1 dependent so I'm not really worried about bumping the major version.
We can also avoid adding a namespace for the request options, a `timeout` option is perfectly fine.

SEAPUNK:
> We can also avoid adding a namespace for the request options, a timeout option is perfectly fine.

+0, I'm a bit leery having _just_ the `timeout` option, considering the only place it's affected is `got`, but at the same time, I don't really have a use case for being able to control all of the options sent to the request function.

If we're adding a default timeout, what should it be? I vote for 1 minute.

lpinca:
[...]

PR #98, opened 2016-08-05
[minor] options.timeout, and request timeouts (default 60000ms)

SEAPUNK:
**do not merge until #97 is merged**

What this times out:

- [x] Time to socket open
- [x] Time to first byte

What this does _not_ time out:

- [ ] Time to last byte (sindresorhus/got#161)

Ref: #96

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,6 +2,7 @@
 
 const camelCase = require('lodash/camelCase');
 const assign = require('lodash/assign');
+const defaults = require('lodash/defaults');
 const path = require('path');
 const got = require('got');
 const fs = require('fs');
@@ -16,12 +17,19 @@ const pkg = require('./package');
  * @param {String} options.apiKey The API Key
  * @param {String} options.password The private app password
  * @param {String} options.accessToken The persistent OAuth public app token
+ * @param {Number} options.timeout=60000 How long to wait in milliseconds
+ *                                       before timing out a request
  * @constructor
  * @public
  */
 function Shopify(options) {
   if (!(this instanceof Shopify)) return new Shopify(options);
-  if (!options || !options.shopName) {
+
+  options = this.options = defaults(options, {
+    timeout: 60000
+  });
+
+  if (!options.shopName) {
     throw new Error('Missing required shopName option');
   }
 
@@ -33,9 +41,7 @@ function Shopify(options) {
   //
   if (options.apiKey && options.password) {
     auth = `${options.apiKey}:${options.password}`;
-  } else if (options.accessToken) {
-    this.token = options.accessToken;
-  } else {
+  } else if (!options.accessToken) {
     throw new Error('Missing required options');
   }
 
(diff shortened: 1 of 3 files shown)


SEAPUNK:
Regarding the time to last byte timeout, do you want me to implement a timeout feature for that? It will be something along the lines of `Promise.race(request, setTimeout(throwError, timeout))`.

SEAPUNK:
Ack, I can't git.

SEAPUNK:
OH! There we go. I almost have 0 experience with rebasing.

lpinca:
> Regarding the time to last byte timeout, do you want me to make the timeout handle that?

No, that's a different issue.

test/shopify.test.js
```diff
@@ -276,5 +276,57 @@ describe('Shopify', () => {
       return shopify.request(url, 'GET')
         .then(res => expect(res).to.deep.equal({}));
     });
+
+    it('times out requests (socket open)', () => {
+      const shopify = new Shopify({
+        shopName: shopName,
```
lpinca: Use property value shorthand.

SEAPUNK: Right.

test/shopify.test.js
```diff
@@ -276,5 +276,57 @@ describe('Shopify', () => {
...
+        apiKey: apiKey,
+        password: password,
+        timeout: 500
+      });
+
+      const url = assign({ path: '/test' }, shopify.baseUrl);
```
lpinca: You can probably reuse the `url` defined in the other scope.

test/shopify.test.js
```diff
@@ -276,5 +276,57 @@ describe('Shopify', () => {
...
+        .socketDelay(1000)
+        .reply(200, {});
+
+      return shopify.request(url, 'GET')
+        .then(() => {
+          throw new Error('Request did not fail');
+        })
+        .catch((err) => {
+          expect(err.message).to.equal(
+            'Socket timed out on request to undefined'
```
lpinca: Do you know why the message includes undefined?

SEAPUNK: It's in the `timed-out` module. Not 100% sure why, but it has to do with how `got` creates requests.

SEAPUNK: https://github.com/floatdrop/timed-out/blob/master/index.js#L6

lpinca: Ok, I guess the `req._headers.host` property does not exists, see https://github.com/floatdrop/timed-out/blob/ead538ea1b5a11c979d8632276b05a36859dbd76/index.js#L19, still curious though.

index.js
```diff
@@ -48,6 +50,11 @@ function Shopify(options) {
     max: undefined
   };
 
+  //
+  // Request timeout.
+  //
+  this.requestTimeout = options.timeout || 60000;
```
lpinca: I wonder if it's better to add a reference to the options object instead of copying the properties to the instance.
So we can use `this.options.timeout` and `this.options.accessToken` below.

SEAPUNK: Yeah, I think it is. Shall I make this change in the PR?

lpinca: Yes that's fine. Also we can use lodash defaults for the defaults options.

SEAPUNK: Changed.

lpinca:
Thank you. I'll fix few nits and release 2.0.0 soon.

SEAPUNK:
Okay, thank you!

PR #103, opened 2016-09-15
[api] Add ApiPermission resource, and ability to uninstall application/revoke OAuth token

carl:


lpinca:
Hmm I'm not too happy with this. The code looks good, but the use case seems edgy and the "ApiPermission" resource is a special one, it has only one action (delete) and is not listed under https://help.shopify.com/api/reference.

As bad as it is I'm more inclined to say use the private `request` method to do this.

```js
const url = Object.assign({
  path: '/admin/api_permissions/current.json'
}, shopify.baseUrl);

shopify.request(url, 'DELETE')
  .then(data => console.log(data))
  .catch(err => console.error(err));
```

benbuckman:
@lpinca Why is this "edgy"? It's one of their API endpoints, documented here: https://help.shopify.com/api/guides/authentication/uninstalling-applications
(I don't know why Shopify neglected to document it on their other page as well, but that's not really important.)

Any app that collects Shopify user tokens and makes API requests needs to be able to revoke those tokens. This is a critical use case.

lpinca:
Their own "official" Ruby [library](https://github.com/Shopify/shopify_api) don't have it, same is for the Python version. Let me think a bit about it.

test/api-permission.test.js
```diff
@@ -0,0 +1,22 @@
...
+
+  const shopify = common.shopify;
+  const scope = common.scope;
+
+  afterEach(() => expect(scope.isDone()).to.be.true);
+
+  it('deletes an api permission', () => {
+    scope
+      .delete('/admin/api_permissions/current.json')
+      .reply(200, {});
```
lpinca: Is this the real thing that is returned? Can't see the response from the example in the doc.

lpinca:
Merged as 9063c0c7991f0036700356838b5f5d5cac1a5f05. Thanks.

PR #105, opened 2016-11-14
Added Product Metafield API

parrabolic:
Added the ability to perform standard operations on Shopify Product Metafields.

Shopify API documentation describing these features: https://help.shopify.com/api/reference/metafield

### DIFF
--- a/resources/product-metafield.js
+++ b/resources/product-metafield.js
@@ -1,0 +1,24 @@
+'use strict';
+
+const assign = require('lodash/assign');
+
+const baseChild = require('../mixins/base-child');
+
+/**
+ * Creates a ProductMetafield instance.
+ *
+ * @param {Shopify} shopify Reference to the Shopify instance
+ * @constructor
+ * @public
+ */
+function ProductMetafield(shopify) {
+  this.shopify = shopify;
+
+  this.parentName = 'products';
+  this.name = 'metafields';
+  this.key = 'metafield';
+}
+
+assign(ProductMetafield.prototype, baseChild);
+
+module.exports = ProductMetafield;


lpinca:
Thanks but I'm going to pass on this.

Metafields can also be added to:

- Blogs
- Custom Collections
- Customers
- Orders
- Pages
- Product Variants

The existing API should already allow you to work with metafields. For example:

Get metafields that belong to a product:

```js
const promise = shopify.metafield.list({
  metafield: { owner_resource: 'product', owner_id: 33411230 }
});
```

Add a metafield to an existing product:

```js
const promise = shopify.product.update(632910392, {
  id: 632910392,
  metafields: [
    {
      key: 'new',
      value: 'newvalue',
      value_type: 'string',
      namespace: 'global'
    }
  ]
});
```

Update a metafield:

```js
const promise = shopify.metafield.update(721389482, {
  id: 721389482,
  value: 'something new',
  value_type: 'string'
});
```

Delete a metafield:

```js
const promise = shopify.metafield.delete(721389482);
```

parrabolic:
The reason for creating a product metafield using the `POST /admin/products/#{id}/metafields.json` url is that the id (and more data) of the metafield is returned in the response. In our use case, this data is then saved so that the same metafield can be updated at a later date.

Using `shopify.product.update` to create metafields returns the **product** which was updated - giving no way to determine the id of the newly created metafield. Using `shopify.product.update` is it possible to retrieve the id of the newly created metafield?

lpinca:
@parrabolic good point I didn't thought about that. It seems that the id of the newly created metafield is not returned when added via `shopify.product.update`.

Is it the the only case where the current API is lacking? If so I would just add `shopify.<resource>.addMetafield`.

Do you know if `shopify.metafield.create` works if you specify an `owner_resource` and `owner_id`? For example

```js
const promise= shopify.metafield.create({
  key: 'warehouse',
  value: 25,
  value_type: 'integer',
  namespace: 'inventory',
  owner_resource: 'product',
  owner_id: 632910392
});
```

parrabolic:
Ah nice, looks like using `shopify.metafield.create` with `owner_resource` and `owner_id` works as expected - returning the created `metafield` in the response.

This should work well enough for my purpose, thanks.

PR #113, opened 2017-01-30
NEEDS FEEDBACK: add metafields for product and variant

kurtharriger:
I need the metafields from product and variants.  

The /admin/products/#{id}/metafields.json endpoint is documented at  https://help.shopify.com/api/reference/metafield rather than the product reference.  

Since it lives under the product endpoint I added it to the product prototype rather than metafields.  Not sure if this is the way you would want to do it, but it seemed the simplest thing that works for me.

### DIFF
--- a/resources/product-variant.js
+++ b/resources/product-variant.js
@@ -49,16 +49,4 @@ ProductVariant.prototype.update = function update(id, params) {
   return this.shopify.request(url, 'PUT', this.key, params);
 };
 
-/**
- * Get all metafields that belong to a product variant
- *
- * @param {Number} variant Variant ID
- * @return {Promise} Promise that resolves with the result
- * @public
- */
-ProductVariant.prototype.metafields = function(id) {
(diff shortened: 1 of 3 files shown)


lpinca:
@kurtharriger thanks but I think I'm going to pass on this.
The existing API should allow you to work with metafields. Take a look at the discussion in this PR https://github.com/MONEI/Shopify-api-node/pull/105.

kurtharriger:
I didn't know you could query on `owner_resource` and  `owner_id` as mentioned in #105
```
shopify.metafield.list({
  metafield: { owner_resource: 'variant', owner_id: 623199805 }
}).then(...);;
```
That seems like a more general solution then this hack.  Thanks!

lpinca:
Yeah this is a recurring issue. I think it makes sense to document how to use metafields with this lib.
Would you mind sending a PR for that?

lpinca:
@Enelar https://github.com/MONEI/Shopify-api-node#metafields

PR #115, opened 2017-02-27
Metafieldcreate

BarryCarlyon:
Adds a call base-child.js for creating a meta field on a product or product variant. (Updated the readme to match)

The same call to createMetafield will also update the same metafield of the matching namespace/key.

I imagine it would would with anything that can have a metafield.

It uses the (main docs) undocumented API end points discussed [here](https://ecommerce.shopify.com/c/shopify-apis-and-technology/t/updating-product-variant-metafields-166055) by Shopify staff

### DIFF
--- a/mixins/base-child.js
+++ b/mixins/base-child.js
@@ -39,6 +39,7 @@ const baseChild = {
 
   /**
    * Creates a new metafield on a record.
+   * This also works to update the existing metafield
    *
    * @param {Number} parentId Parent record ID
    * @param {Object} params Record properties
--- a/mixins/base-child.js
+++ b/mixins/base-child.js
@@ -36,13 +36,22 @@ const baseChild = {
     const url = this.buildUrl(parentId);
     return this.shopify.request(url, 'POST', this.key, params);
   },
+
+  /**
+   * Creates a new metafield on a record.
+   *
+   * @param {Number} parentId Parent record ID
+   * @param {Object} params Record properties
+   * @return {Promise} Promise that resolves with the result
+   * @public
+   */
   createMetafield(parentId, params) {
-    // rearrange arguments
+    // rearrange and override arguments
     this.parentName = this.name;
-    this.name = '';
+    this.name = 'metafield';
     this.key = 'metafield';
 
-    const url = this.buildUrl(parentId, 'metafield');
+    const url = this.buildUrl(parentId);
     return this.shopify.request(url, 'POST', this.key, params);
   },
 
--- a/mixins/base-child.js
+++ b/mixins/base-child.js
@@ -36,6 +36,15 @@ const baseChild = {
     const url = this.buildUrl(parentId);
     return this.shopify.request(url, 'POST', this.key, params);
   },
+  createMetafield(parentId, params) {
+    // rearrange arguments
+    this.parentName = this.name;
+    this.name = '';
+    this.key = 'metafield';
+
+    const url = this.buildUrl(parentId, 'metafield');
+    return this.shopify.request(url, 'POST', this.key, params);
+  },
 
   /**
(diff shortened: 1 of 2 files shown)


lpinca:
Hi,
take a look at the discussion in #105.

BarryCarlyon:
Heh I missed that closed PR

The problem with shopify.product.update is that it will fail if the metafield already exists, and the response from shopify.product.update doesn't return the metafield itself.

It makes no sense for anyone to use shopify.product.update to update/first set the metafield since you have to do another API query, to _get_ the metafield

You are forcing people to waste API requests, by not offering the standalone metafield creation, since the standalone method right now will create or update it and return it.

(Shopify is a utter mess when it comes to Metafields, so this PR gives flexibility)

*shrug* I'll just use my own fork if I have to. This method works better for my use case.

lpinca:
> The problem with shopify.product.update is that it will fail if the metafield already exists, and the response from shopify.product.update doesn't return the metafield itself.

I think you only read part of the discussion. See https://github.com/MONEI/Shopify-api-node/pull/105#issuecomment-260283314

```js
const promise = shopify.metafield.create({
  key: 'warehouse',
  value: 25,
  value_type: 'integer',
  namespace: 'inventory',
  owner_resource: 'product',
  owner_id: 632910392
});
```

This allows you to create a metafield for a product and get back the created metafield.

BarryCarlyon:
Which then means you need a local record, or need to do a look up to see if it exists first.

 - Need to update field
 - Check field exists
 - Create or update

This PR adds the function that needs neither.

 - Need to update field
 - Just send the metafield

*sigh* shopify's API is just horrid

I'll revisit my flows, but right now I'm using the POST /admin/variants/$PRODUCT_VARIANT_ID/metafields.json

lpinca:
> I'll revisit my flows, but right now I'm using the POST /admin/variants/$PRODUCT_VARIANT_ID/metafields.json

I'm confident that the result is exactly the same when using `shopify.metafield.create()` with the `owner_resource` and `owner_id` params.

Try that.

BarryCarlyon:
It may be so.

But shouldn't this "SDK" be not opinionated and offer all the ability to call any of the api endpoints shopify providers?

lpinca:
We can add `shopify.<resource>.metafield.<method>` for resources that allow metafields but that is uncessary complexity imo as you can have the same functionality with just 2 params `owner_resource` and `owner_id` which are also documented by Shopify.

mixins/base-child.js
```diff
@@ -38,6 +38,25 @@ const baseChild = {
...
   /**
+   * Creates a new metafield on a record.
+   * This also works to update the existing metafield
+   *
+   * @param {Number} parentId Parent record ID
+   * @param {Object} params Record properties
+   * @return {Promise} Promise that resolves with the result
+   * @public
+   */
+  createMetafield(parentId, params) {
```
lpinca: `Product` does not use this mixin so this method will not be available for `shopify.product`.

mixins/base-child.js
```diff
@@ -38,6 +38,25 @@ const baseChild = {
...
+    // rearrange and override arguments
+    this.parentName = this.name;
+    this.name = 'metafield';
```
lpinca: You are changing "private" properties of the instance with these assignments.
What happens if you try to use say `shopify.productVariant.get(parentId)` after calling `shopify.productVariant.createMetafield(parentId, params)`?

lpinca:
> But shouldn't this "SDK" be not opinionated and offer all the ability to call any of the api endpoints shopify providers?

You can do that using the low level `shopify.request()`.

BarryCarlyon:
Makes sense. Gonna dig about somemore

PR #116, opened 2017-03-27
Implement TokenBucket instead of valvelet

BarryCarlyon:
Shopify uses a leaky bucket algo for it's API which means it supports a spike in API usage of up to 40 requests refilling at 2 per second.

I couldn't figure out why my usage of this module wasn't implementing this, so I took a look under the hood and found `valvelet` was in use. After some digging I found `TokenBucket`. And with testing, capped it to 38 requests.

Now when you dump a tonne of requests onto this Shopify, it'll use 38 of 40 requests and then 2 per second until the queue is complete.

You can at boot configure the three parameters as needed also, (noted in the README)

I also pass the number of `TokenBucket` tokens left to the updateLimits functions to expose it for use/checking.

In Summary:

> This PR switches from `valvelet` to `TokenBucket` and revises the README appropriately.


lpinca:
It makes sense but this PR can't be merged in its current state.

- The functionality should be opt-in (disabled by default) as it is now. Users may want to use their own strategy to rate limit requests.
- Failing tests should be fixed and new ones added.
- I would like to avoid bumping the major version just for this enhancement so new options can be added but existing ones can't be removed or changed.
- I don't like `TokenBucket`. It has a lot of features which are useless here. Is there a simpler module? Consider sending a PR to valvelet, I would be happy to merge a PR there which adds "burst" support.

On another topic, why you want to keep the bucket empty? It saves some time if your app does infrequent bursts but is it really needed?

BarryCarlyon:
I'm sending stock updates to shopify as and when they occur.

For some time periods nothing happens, then I get an delivery from a supplier and boom I've got a ton of products to update. So spike to burst rate to smooth the ramp up, which for small suppliers that deliver less than 40 items, boom I'm done and then back to order processing as normal, or regular stock updates. So right now I'm bursting all the time anyway.

I'll look at and address the other issues.

BarryCarlyon:
I also have no idea why the travis build failed. The job log won't load for me

lpinca:
Run `npm test` from your local fork, you should get the same failures.

BarryCarlyon:
roger that I see some fails to go and fix/test for now

lpinca:
I didn't a valid alternative for `valvelet`, so I wrote [`stopcock`](https://github.com/lpinca/stopcock) which works like `valvelet` but limits the execution rate of a function using the leaky bucket algoritm.

Would you mind testing it? It is possible that there are bugs or it doesn't works as expected.

Thanks.

BarryCarlyon:
I'll give it a whirl when I get a sec

PR #117, opened 2017-03-28
Apply an EventEmitter to updateLimits

BarryCarlyon:
Adds an EventEmitter to the updateLimits function, so we can obtain limits in real time instead of long polling, for monitoring etc

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,13 @@ is an object like this:
 }
 ```
 
-Values start at `undefined` and are updated every time a request is made.
+Values start at `undefined` and are updated every time a request is made. After
+every update the `callLimits` event is emitted with the updated limits as
+argument.
+
+```js
+shopify.on('callLimits', limits => console.log(limits));
+```
 
 ### Resources
 
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 const camelCase = require('lodash/camelCase');
 const defaults = require('lodash/defaults');
 const assign = require('lodash/assign');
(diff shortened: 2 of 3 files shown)


lpinca:
This would be the only event so it's probably better to just add an hook.

```js
shopify.onUpdateLimits = function (limits) {...};
```

and call this from `shopify.updateLimits()` if present.

BarryCarlyon:
So something like:


```
/**
 * Updates API call limits.
 *
 * @param {String} header X-Shopify-Shop-Api-Call-Limit header
 * @private
 */
Shopify.prototype.updateLimits = function updateLimits(header, remainingTokens) {
  if (!header) return;

  const limits = header.split('/').map(Number);
  const callLimits = this.callLimits;

  callLimits.remaining = limits[1] - limits[0];
  callLimits.current = limits[0];
  callLimits.max = limits[1];

  if (remainingTokens) {
    callLimits.internal = Math.floor(remainingTokens);
  }

  if (this.onUpdateLimits) {
    this.onUpdateLimits(callLimits);
  }
};
Shopify.prototype.onUpdateLimits = false;
```

lpinca:
Yes, without adding the prototype property. We can set it in the constructor.

```js
this.onUpdateLimits = null;
```

If you feel strong about making the `Shopify` class an `EventEmitter` because you think it might be useful to have more events in future please use https://github.com/primus/eventemitter3.

The advantage is that it works in all browsers (some people "browserify" this lib).

PR #129, opened 2017-06-23
Add in the sendInvite API endpoint

BarryCarlyon:
Working on customer stuff at $dayjob again

Testing with activations and spotted that the sendInvite API call is missing from the module. Added the call

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,7 @@ shopify.metafield.create({
   - `update(id, params)`
 - customer
   - `accountActivationUrl(id)`
+  - `sendInvite(id[, params])`
   - `count()`
   - `create(params)`
   - `delete(id)`
--- a/README.md
+++ b/README.md
@@ -236,13 +236,13 @@ shopify.metafield.create({
   - `update(id, params)`
 - customer
   - `accountActivationUrl(id)`
-  - `sendInvite(id[, params])`
   - `count()`
   - `create(params)`
   - `delete(id)`
   - `get(id[, params])`
   - `list([params])`
   - `search(params)`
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -236,6 +236,7 @@ shopify.metafield.create({
   - `update(id, params)`
 - customer
   - `accountActivationUrl(id)`
+  - `sendInvite(id[, params])`
```
lpinca: Please keep it sorted.

resources/customer.js
```diff
@@ -46,4 +46,17 @@ Customer.prototype.accountActivationUrl = function accountActivationUrl(id) {
...
+ * Generates a invite email for a customer.
+ *
+ * @param {Number} id Customer ID
+ * @param {Object} params Optional params for adjusting the sent email
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+Customer.prototype.sendInvite = function sendInvite(id, params) {
+  const url = this.buildUrl(`${id}/send_invite`);
+  return this.shopify.request(url, 'POST', 'customer_invite', params).then(body => body.customer_invite);
```
lpinca: This should be

```js
return this.shopify.request(url, 'POST', 'customer_invite', params || {});
```

or it won't work with default invite.
The `then()` part is wrong as the promise is already resolved with ` body[key]` if the `key` argument is used in `Shopify.prototype.request()`.

lpinca:
We also need tests.

BarryCarlyon:
Derp forgot the tests will take a look on Monday…

lpinca:
Thanks.

PR #134, opened 2017-08-14
Add refund calculation endpoint

builtbylane:


resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
 
 assign(Refund.prototype, pick(baseChild, ['get', 'buildUrl']));
 
+
+/**
+ * Calculates a refund.
+ * https://help.shopify.com/api/reference/refund#calculate
```
lpinca: I'd remove this line as this is the only case where it is used.

resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
...
+ *
+ * @param {Number} parentId Parent record ID
+ * @param {Object} [params] Query parameters
```
lpinca: This argument is not needed, please remove it.

resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
...
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+Refund.prototype.calculate = function complete(orderId, params) {
```
lpinca: Ditto, please remove the `params` argument.

resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
...
+  const url = this.buildUrl(orderId, '/calculate');
```
lpinca: Can you please remove the `/` from `'/calculate'`? It doesn't cause any harm but it's redundant.

resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
...
+  return this.shopify.request(url, 'POST', undefined, params)
```
lpinca: Use `this.key` as third argument, remove `params`.

resources/refund.js
```diff
@@ -22,4 +22,20 @@ function Refund(shopify) {
...
+    .then(body => body[this.key]);
```
lpinca: Remove this line, it's not needed if you add `this.key` as suggested in the previous comment.

lpinca:
Thanks.
There is still some work to do:

1. Add tests.
2. Update the documentation.

lpinca:
Merged as e6bb4f9596286ab7b8652268b7e70e6ca9731f53.
Thank you.

builtbylane:
thanks for cleaning this up.

PR #136, opened 2017-08-17
add a create refund endpoint

builtbylane:
using the update from the basechild did not work, so had to manually add the update function.

### DIFF
--- a/resources/refund.js
+++ b/resources/refund.js
@@ -35,4 +35,17 @@ Refund.prototype.calculate = function calculate(orderId, params) {
   return this.shopify.request(url, 'POST', this.key, params);
 };
 
+/**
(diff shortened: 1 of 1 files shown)


lpinca:
Hi,
this has been added in https://github.com/MONEI/Shopify-api-node/commit/2af36708332539bfadf0fd74953b188bdc3bc17a

lpinca:
The update action is no documented https://help.shopify.com/api/reference/refund so it has been omitted https://github.com/MONEI/Shopify-api-node/blob/8523ebf04689bbf41cb3433beb572e35b77f9855/resources/refund.js#L23.

builtbylane:
yeah, you are correct. sorry about that.

PR #146, opened 2017-11-16
Adding in missing PUT and DELETE for productListings

bparise:
This change adds in PUT (via create) and DELETE for productListings endpoint

### DIFF
--- a/resources/product-listing.js
+++ b/resources/product-listing.js
@@ -20,10 +20,34 @@ function ProductListing(shopify) {
 }
 
 assign(ProductListing.prototype, omit(base, [
-  'create',
-  'delete',
   'update'
 ]));
+
+/**
+ * Adds the productId to a sales channel
+ *
+ * @param {Number} productId ProductID to add to the sales channel
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+ProductListing.prototype.create = function addProductId(productId) {
+  const url = this.buildUrl(`${productId}`);
+  return this.shopify.request(url, 'PUT', undefined, {})
+    .then(body => body[this.key]);
+};
+
+/**
+ * Deletes the productId to a sales channel
+ *
+ * @param {Number} productId ProductID to delete from the sales channel
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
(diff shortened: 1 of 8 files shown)


bparise:
Dope. Forgot test cases to cover this pull request... will do shortly

resources/product-listing.js
```diff
@@ -20,11 +20,35 @@ function ProductListing(shopify) {
...
+/**
+ * Adds the productId to a sales channel
+ *
+ * @param {Number} productId ProductID to add to the sales channel
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+ProductListing.prototype.create = function addProductId(productId) {
+  const url = this.buildUrl(`${productId}`);
+  return this.shopify.request(url, 'PUT', undefined, {})
```
lpinca: The [example](https://help.shopify.com/api/reference/productlisting#create) does not use an empty body. Can we do something like this?

```js
/**
 * Creates a product listing.
 *
 * @param {Number} productId The ID of the product to publish
 * @param {Object} [params] Body parameters
 * @return {Promise} Promise that resolves with the result
 * @public
 */
ProductListing.prototype.create = function create(productId, params) {
  params || (params = { product_id: productId });
  const url = this.buildUrl(productId);
  return this.shopify.request(url, 'PUT', this.key, params || { product_id: productId });
};
```

resources/product-listing.js
```diff
@@ -20,11 +20,35 @@ function ProductListing(shopify) {
...
+};
+
+/**
+ * Deletes the productId to a sales channel
+ *
+ * @param {Number} productId ProductID to delete from the sales channel
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+ProductListing.prototype.delete = function deleteProductId(productId) {
```
lpinca: I think we can remove this as the default implementation https://github.com/MONEI/Shopify-api-node/blob/53b6ac8c1253a46aa423f415bb0c3f7bdd80df95/mixins/base.js#L45-L48 should work.

lpinca:
We also need to update the documentation.

lpinca:
Thank you.

Issue #149, opened 2017-12-04
An interface would be amazing


stephengardner:
The documentation around this API was actually confusingly difficult to identify.  I couldn't tell what a Shopify Admin App exactly was.  There's public, private apps.  All sorts of stuff.  Anyways, I digress.

Going here https://help.shopify.com/api/reference and then clicking on an endpoint and then clicking again to find the return value of the method in question and then coming back here to see the correlating function to use on this module is a bit of a relative hassle in the age of excellent developer docs.  If this module had a Typescript interface that would make my day.

Thanks for the module.

lpinca:
Any kind of help is appreciated. I don't use TypeScript so I can't write/maintain definitions. You could write and submit them to https://github.com/DefinitelyTyped/DefinitelyTyped if you want. Code is documented with JSDoc so it shouldn't be very hard.

buggy:
I have partial TypeScript definitions file that I'm happy to contribute and help maintain. It's a start but it's missing a lot of API's. If this is going to be the official Shopify Node SDK then ideally we could include it here rather than DefinitelyTyped. If you're interested I can put together a PR.

lpinca:
@buggy I'm fine with it, I just don't to want to have a broken/outdated definitions file.

I don't think this is the "official" library as it is still listed under "third-party" in their documentation https://help.shopify.com/api/supported-libraries. @alexandresaiz should have more info on this.

stephengardner:
Yesterday I started a very basic amalgamation of some type rules and published it.  I'll drop the updated link here shortly.

The repo README says "Official module", so I wasn't sure if it was just semantics lol.  Cool, thanks

stephengardner:
I've given the top level module a declaration, plugged every action with at least an "any" type and begun working on a few actions like `product`.
https://github.com/stephengardner/shopify-api-node-interface

buggy:
I want to provide a quick update so everyone knows progress is being made. Late last week I scraped the API docs from the Shopify website to generate interfaces for the API objects. Unfortunately the API docs rarely seem to match the actual API and that is slowing down definition creation.

Given the problems with the API docs the only way we'll know for sure if the definitions are correct is when people start using them.

I'm only able to work on this in my spare time so it will probably take another 2-3 weeks to create all of the definitions. My goal is to have the first PR ready by the end of this week then a new PR each week as more definitions are added.

buggy:
[...]

PR #150, opened 2017-12-08
JSDoc improvements

dhritzkiv:
Using TypeScript's checkJS functionality in Visual Studio Code, my constructor options were often marked as incorrect.

This will correctly allow IDEs to hint at which properties are allowed.

Unfortunately, I wasn't able to create mutually exclusive definitions of which options are allowed. Which means, at code-time, passing in both `accessToken` and `password` will be marked as correct (at runtime, however, an error will be thrown).

Still, this is an improvement over valid options being marked as invalid.

### DIFF
--- a/index.js
+++ b/index.js
@@ -13,15 +13,30 @@ const fs = require('fs');
 const pkg = require('./package');
 
 /**
+ * @typedef {object} ConstructorOptionsAlways
+ * @property {string} shopName
+ * @property {string} [apiKey]
+ * @property {string} [password]
+ * @property {string} [accessToken]
+ * @property {boolean | object} [autoLimit]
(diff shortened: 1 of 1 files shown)


lpinca:
Technically it looks good but it seems unnecessarily complex. With TypeScript definitions being work on (https://github.com/MONEI/Shopify-api-node/issues/149) I don't know if it makes much sense.

dhritzkiv:
Ah, that's cool. I didn't realize there were typings being worked on. I figured a quick patch to the JSDoc  approach would be beneficial, but obviously typings are the more perfect solution.

lpinca:
Thank you anyway.

PR #155, opened 2018-01-10
[ts] Initial commit of TypeScript definitions

buggy:
This PR is a first attempt at the TypeScript definitions. It has been scraped together from the API documentation and some experimentation with the actual API. 

It includes:

* A definition for the Shopify class as well as definitions for public and private app options passed to it.
* All of the API call are defined. Where types haven't been defined `any` is used.
* Most of the objects returned by the Shopify API are defined
* Some of the create/update API calls also have definitions for their parameters

@lpinca If you're happy to include this in the main project then I'm willing to maintain it.

@stephengardner I'd appreciate a second set of eyes over the definition if you have time.

Moving forward the plan is to expand the definitions to additional API's as required or as time permits.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1987 @@
+// Type definitions for Shopify-api-node
+// Project: Shopify-api-node
+// Definitions by: Rich Buggy <rich@buggy.id.au>
+
+/*~ This is the module template file for class modules.
+ *~ You should rename it to index.d.ts and place it in a folder with the same name as the module.
+ *~ For example, if you were writing a file for "super-greeter", this
+ *~ file should be 'super-greeter/index.d.ts'
+ */
+
+/*~ Note that ES6 modules cannot directly export class objects.
+ *~ This file should be imported using the CommonJS-style:
+ *~   import x = require('someLibrary');
+ *~
+ *~ Refer to the documentation to understand common
+ *~ workarounds for this limitation of ES6 modules.
+ */
+
+/*~ This declaration specifies that the class constructor function
+ *~ is the exported object from the file
+ */
+export = Shopify;
+
+/*~ Write your module's methods and properties in this class */
+declare class Shopify {
+    constructor(config: Shopify.IPublicShopifyConfig | Shopify.IPrivateShopifyConfig);
+    callLimits: Shopify.ICallLimits;
+    apiPermission: {
+        delete(): void;
+    };
+    // abandonedCheckouts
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,1987 @@
...
+    payment: {
+        count: (checkoutToken: any) => Promise<number>;
+        create: (checkoutToken: any, params: any) => Promise<any>;
+        get: (checkoutToken: any, id: number) => Promise<any>;
+        list: (checkoutToken?: any) => Promise<any>;
+    };
+    policy: {
+        list: (params?: any) => Promise<Shopify.IPolicy[]>;
+    };
+    priceRules: {
```
lpinca: If I'm not wrong this should be `priceRule`.

index.d.ts
```diff
@@ -0,0 +1,1987 @@
...
+        "themes/delete" |
+        "themes/publish" |
+        "themes/update";
+
+    type WebhookFormat = "json" | "xml";
+
+    export interface IWebhook {
+        address: string;
+        created_at: string;
+        fields: s
```
lpinca: Please add a new line.

index.d.ts
```diff
@@ -0,0 +1,1987 @@
...
+    interface IImage {
+        created_at: string;
+        height: number;
+        src: string;
+        updated_at?: string;
+        width: number;
+    }
+
+    interface IBase64Image {
+        attrachment: string;
```
lpinca: `attachment` ?

lpinca:
Can you please include this in the [`files`](https://github.com/MONEI/Shopify-api-node/blob/2d372bd2381d8ad59df6fbd602c43b353f285a91/package.json#L12-L16) array?

lpinca:
> @lpinca If you're happy to include this in the main project then I'm willing to maintain it.

Yes that would be great. I wonder if @alexandresaiz is ok with adding you as collaborator to the project.

index.d.ts
lpinca: Not sure why but this comment seems misplaced, I meant new line at the end of file.

alexandresaiz:
@lpinca fyi, @buggy has been added with write perm.

lpinca:
Awesome, thank you.

lpinca:
I'll keep this open for a couple of days in case anyone want to take a look.

lpinca:
Released version 2.10.1, thank you @buggy.

PR #160, opened 2018-01-17
[ts] Improve constructor option definitions

dhritzkiv:
- marks timeout as an optional field
- use and extend a common interface for both config interfaces

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -408,7 +408,7 @@ declare namespace Shopify {
         accessToken: string;
         autoLimit?: boolean | IAutoLimit;
         shopName: string;
-        timeout: number;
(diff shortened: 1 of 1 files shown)


lpinca:
cc: @buggy

buggy:
@dhritzkiv Can you remove the second commit from this PR (the one that creates ICommonShopifyConfig)? There are only 2 common fields and having this makes the code less readable.

Also, can you rename the PR to start with [ts] so it's easy to track all of the TypeScript commits.

dhritzkiv:
Removed the second commit

lpinca:
Thank you.

Issue #170, opened 2018-03-11
New Access Scope API


buggy:
I notice Shopify have added a new [Access Scope API](https://help.shopify.com/api/reference/accessscope). We should add support for it.

PR #171, opened 2018-03-12
Add AccessScope resource

lpinca:
Fixes: https://github.com/MONEI/Shopify-api-node/issues/170

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -27,8 +27,11 @@ type onCallLimitsFn = (limits: Shopify.ICallLimits) => void;
 declare class Shopify {
     constructor(config: Shopify.IPublicShopifyConfig | Shopify.IPrivateShopifyConfig);
     callLimits: Shopify.ICallLimits;
+    accessScope: {
+        list: () => Promise<Shopify.IAccessScope[]>;
+    };
     apiPermission: {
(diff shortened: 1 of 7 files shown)


index.d.ts
```diff
@@ -27,8 +27,11 @@ type onCallLimitsFn = (limits: Shopify.ICallLimits) => void;
 declare class Shopify {
     constructor(config: Shopify.IPublicShopifyConfig | Shopify.IPrivateShopifyConfig);
     callLimits: Shopify.ICallLimits;
+    accessScope: {
+        list: () => Promise<Shopify.IAccessScope[]>
```
lpinca: @buggy is this ok?

index.d.ts
```diff
@@ -428,6 +431,10 @@ declare namespace Shopify {
         max: number;
     }
 
+    interface IAccessScope {
+        handle: string
```
lpinca: I think we can define a type with all [scopes](https://help.shopify.com/api/getting-started/authentication/oauth/scopes) and use it here instead of just `string` but I'm not sure if it makes sense.

lpinca:
Ping @buggy. Can you please review the TypeScript changes when you have time?
Thank you.

Issue #176, opened 2018-04-09
Implement InventoryLevel


EvanDarwin:
https://help.shopify.com/api/reference/inventorylevel

For ProductVariants, `inventory_quantity`, and `old_inventory_quantity` are deprecated and cannot be updated anymore.

PR #177, opened 2018-04-10
Add InventoryLevel resource

lpinca:
Fixes: https://github.com/MONEI/Shopify-api-node/issues/176

cc: @buggy for TypeScript type  definitions.


lpinca:
If anyone is interested in TypeScript type definitions for this, please open a PR.

PR #180, opened 2018-04-12
[ts] Fixed Typo - IProudctOptions => IProductOptions

EvanDarwin:
There is a typo in the TS definition, this should be a subtle change. Unsure if you want to introduce some kind of backwards compatibility for `IProudctOptions` (I chose not to clutter things up).


lpinca:
Thank you.

PR #187, opened 2018-05-17
Update README.md with usage example

hcmlopes:
Adds example to discountCode.create to show that it does not need the full "discount_code" object as the param as shown in the shopify API docs


lpinca:
This is valid for all resources not only `DiscoutCode`. I would prefer a paragraph with an example under the [Resources](https://github.com/MONEI/Shopify-api-node#resources) section to make it extra clear.

lpinca:
Thank you.

PR #194, opened 2018-06-26
Added more details for init

kpthree:
Added a message which states what could be the values for shopName at the time of init

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ const shopify = new Shopify({
 
 // shopify.<resource_name>.<method_name>
 ```
-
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -108,7 +108,7 @@ const shopify = new Shopify({
 
 // shopify.<resource_name>.<method_name>
 ```
-
+`shopName` can be with or without `.myshopify.com`. eg `test-shopify.myshopify.com` or `test-shopify`
```
lpinca: Can you please move this in the description of the `shopName` option? https://github.com/MONEI/Shopify-api-node#options

Also please keep the limit of 80 chars per line.
Thanks.

PR #197, opened 2018-07-10
Unwrapping the result of /recurring_application_charges//activate API

arypbatista:
Now the API wrapper is returning the unwrapped value on `recurringBilling.activate` method. 

At some point the right test should be implemented. Right now the test for this seems to be invalid since the result obtained through the test is an empty object and Shopify API's documentation says it should be returning the updated charge object.

[See RecurringApplicationCharge documentation](https://help.shopify.com/en/api/reference/billing/recurringapplicationcharge#activate)

### DIFF
--- a/test/recurring-application-charge.test.js
+++ b/test/recurring-application-charge.test.js
@@ -78,6 +78,15 @@ describe('Shopify#recurringApplicationCharge', () => {
   });
 
   it('activates a recurring application charge', () => {
+    /**
+     * TODO: According to Shopify API the activation of a charge
+     * should return the updated charge object with the 'status'
+     * value set as 'active'. This test seems to be trying to
+     * activate an invalid charge ID, and might be the reason why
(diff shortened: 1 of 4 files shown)


test/recurring-application-charge.test.js
```diff
@@ -78,6 +78,15 @@ describe('Shopify#recurringApplicationCharge', () => {
   });
 
   it('activates a recurring application charge', () => {
+    /**
+     * TODO: According to Shopify API the activation of a charge
+     * should return the updated charge object with the 'status'
+     * value set as 'active'. This test seems to be trying to
```
lpinca: It seems the response in docs i correct now, see https://help.shopify.com/en/api/reference/billing/recurringapplicationcharge#activate. You can take that as fixture.

Thanks.

arypbatista: Roger, will do

lpinca:
Went ahead and pushed on your branch.

Thank you.

PR #202, opened 2018-07-16
Typescript defs

juanifit:


index.d.ts
```diff
@@ -1398,6 +1399,13 @@ declare namespace Shopify {
         refunds: string;
         shipping_address: ICustomerAddress;
         shipping_lines: IOrderShippingLine[];
+        subtotal_price: string;
+        total_discounts: string;
+        total_line_items_price: string;
+        total_price: string;
+        total_tax: string;
+        total_tip_received: string;
+        total_weight: string;
```
lpinca: This is a number.

index.d.ts
```diff
@@ -161,6 +161,7 @@ declare class Shopify {
         enable: (id: number) => Promise<Shopify.IDiscountCode>;
         get: (id: number) => Promise<Shopify.IDiscountCode>;
         list: (params?: any) => Promise<Shopify.IDiscountCode[]>;
+        lookup: (params?: any) => Promise<Shopify.IDiscountCode>;
```
lpinca: `params` is not optional.

index.d.ts
```diff
@@ -161,6 +161,7 @@ declare class Shopify {
         enable: (id: number) => Promise<Shopify.IDiscountCode>;
```
lpinca: Now that you are here, can you also please fix other issues here?

- `create` parameters should be `(priceRuleId: number, params: any)`.
- `delete` parameters should be `(priceRuleId: number, id: number)`.
- `enable` and `disable` do not exist, so they can be removed.
- `get` parameters should be `(priceRuleId: number, id: number)`.
- `list` parameters should be `(priceRuleId: number)`.
- `update` should be added:
  `(priceRuleId: number, id: number, params: any) => Promise<Shopify.IDiscountCode>`.

Thank you.

PR #204, opened 2018-07-20
fix typing for payment and discountCode

mastilver:
Hi

A couple of children resources didn't include the parent id

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -273,10 +273,10 @@ declare class Shopify {
         update: (id: number, params: any) => Promise<Shopify.IPage>;
     };
     payment: {
-        count: (checkoutToken: any) => Promise<number>;
-        create: (checkoutToken: any, params: any) => Promise<any>;
-        get: (checkoutId:number, checkoutToken: any, id: number) => Promise<any>;
-        list: (checkoutToken?: any) => Promise<any>;
+        count: (checkoutToken: string) => Promise<number>;
+        create: (checkoutToken: string, params: any) => Promise<any>;
+        get: (checkoutToken: string, id: number) => Promise<any>;
+        list: (checkoutToken?: string) => Promise<any>;
     };
     policy: {
         list: (params?: any) => Promise<Shopify.IPolicy[]>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -157,11 +157,10 @@ declare class Shopify {
     discountCode: {
         create: (priceRuleId: number, params: any) => Promise<Shopify.IDiscountCode>;
         delete: (priceRuleId: number, id: number) => Promise<void>;
-        disable: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
-        enable: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
         get: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
         list: (priceRuleId: number, params?: any) => Promise<Shopify.IDiscountCode[]>;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -155,12 +155,12 @@ declare class Shopify {
...
     discountCode: {
-        create: (params: any) => Promise<Shopify.IDiscountCode>;
-        delete: (id: number) => Promise<void>;
-        disable: (id: number) => Promise<Shopify.IDiscountCode>;
-        enable: (id: number) => Promise<Shopify.IDiscountCode>;
-        get: (id: number) => Promise<Shopify.IDiscountCode>;
-        list: (params?: any) => Promise<Shopify.IDiscountCode[]>;
+        create: (priceRuleId: number, params: any) => Promise<Shopify.IDiscountCode>;
+        delete: (priceRuleId: number, id: number) => Promise<void>;
+        disable: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
```
lpinca: This does not exist.

index.d.ts
```diff
@@ -155,12 +155,12 @@ declare class Shopify {
...
+        enable: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
```
lpinca: Ditto.

index.d.ts
```diff
@@ -155,12 +155,12 @@ declare class Shopify {
...
+        get: (priceRuleId: number, id: number) => Promise<Shopify.IDiscountCode>;
+        list: (priceRuleId: number, params?: any) => Promise<Shopify.IDiscountCode[]>;
```
lpinca: `lookup` and `update` are missing.

index.d.ts
```diff
@@ -271,10 +271,10 @@ declare class Shopify {
         update: (id: number, params: any) => Promise<Shopify.IPage>;
     };
     payment: {
-        count: (checkoutToken: any) => Promise<number>;
-        create: (checkoutToken: any, params: any) => Promise<any>;
-        get: (checkoutToken: any, id: number) => Promise<any>;
-        list: (checkoutToken?: any) => Promise<any>;
+        count: (checkoutId:number, checkoutToken: any) => Promise<number>;
```
lpinca: No changes are needed here as per https://help.shopify.com/en/api/reference/sales_channels/payment

index.d.ts
```diff
@@ -276,7 +275,7 @@ declare class Shopify {
     payment: {
         count: (checkoutToken: any) => Promise<number>;
         create: (checkoutToken: any, params: any) => Promise<any>;
-        get: (checkoutToken: any, id: number) => Promise<any>;
+        get: (checkoutId:number, checkoutToken: any, id: number) => Promise<any>;
```
lpinca: This doesn't look correct, I think it shouldn't be changed, see https://help.shopify.com/en/api/reference/sales_channels/payment#show

index.d.ts
```diff
@@ -1555,7 +1554,7 @@ declare namespace Shopify {
         name: string;
         value: string
     }
-                              
+
```
lpinca: ❤️

index.d.ts
mastilver: Right, somehow missed the id...

BTW, `checkoutToken` should be a `string`, right or is there a reason it's `any` right now?

lpinca: Yes you are right, I think there is no particular reason, thanks for fixing it.

lpinca:
Thank you.

PR #205, opened 2018-07-20
[ts] Add option_values to variants

squirly:
See the poorly documented `options_values` field that is under View Response in [GET Product Listing](https://help.shopify.com/en/api/reference/sales_channels/productlisting#show).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1509,10 +1509,17 @@ declare namespace Shopify {
         updated_at: string;
     }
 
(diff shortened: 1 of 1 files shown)


lpinca:
Are these also returned when using `productVariant`? I can see them here https://help.shopify.com/en/api/reference/products/product_variant.

squirly:
Odd. It seems to only be available in product listings. I will update these changes to reflect this.

PR #209, opened 2018-07-24
 [ts] Fix TypeScript definition syntax

squirly:
Seems like the last commit on the branch was not included in my last PR.


lpinca:
Thanks.

PR #210, opened 2018-07-26
[ts] fix options type in IProductListing

squirly:
See [Sales Channel: ProductListing#options](https://help.shopify.com/en/api/reference/sales_channels/productlisting#options-property)


lpinca:
Thank you.

Issue #217, opened 2018-08-09
IProductImage with alt attribute


PabloRosales:
Shopify API sends now the alt attribute, would be nice if IProductImage type is updated for this.

IProduct.image is not using IProductImage.

Also IPage has a typo in: template_suffic and is not using IObjectMetafield for metafield.

I can make a pull request for this in the next days if needed.

Thanks for the great lib :)

lpinca:
Yes please do it.

lpinca:
Fixed by #218.

PR #218, opened 2018-08-09
Improving types

PabloRosales:
Hey, adding some improvements on types for #217

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -646,6 +646,7 @@ declare namespace Shopify {
         src: string;
         updated_at?: string;
         width: number;
+        alt?: string | null;
     }
 
     interface IBase64Image {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1516,6 +1517,7 @@ declare namespace Shopify {
         width: number;
         height: number;
         updated_at: string;
+        alt: string | null;
```
lpinca: This is not documented, https://help.shopify.com/en/api/reference/products/product_image is it available?

PabloRosales: It is as far as I can see in all responses (shown in documentation) involving images, did some tests and I did get them when available and as null when not set.

lpinca: Ok, thank you.

lpinca:
Thank you.

PabloRosales:
Thank you @lpinca for all the work here.

PR #219, opened 2018-08-19
Updates typescript definition for Orders to include additional fields

kjaenicke:
### Update TypeScript definitions to adjust fields returned from API for Orders

The following fields were not include in the `IOrder` interface and have been added:

- `token`
- `user_id`
- `updated_at`
- `order_status_url`

The definition had `refunds` listed as a string instead of an array and has been updated.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1398,7 +1398,6 @@ declare namespace Shopify {
         processed_at: string;
         processing_method: OrderProcessingMethod;
         referring_site: string;
-        refunds: string;
         shipping_address: ICustomerAddress;
         shipping_lines: IOrderShippingLine[];
         subtotal_price: string;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1408,6 +1407,22 @@ declare namespace Shopify {
         total_tax: string;
         total_tip_received: string;
         total_weight: number;
+        token: string;
+        user_id: number | null;
+        updated_at: string;
+        order_status_url: string;
+        refunds: {
```
lpinca: Shouldn't this be `IRefund[]`?

kjaenicke: Yes, you're right. Updated.

lpinca:
Thank you.

PR #222, opened 2018-08-29
[ts] add order_status_url to IOrder type

squirly:
Documented in https://help.shopify.com/en/api/reference/orders/order#order-status-url-property


lpinca:
Thank you.

PR #228, opened 2018-09-17
Update customer.js - method orders to allow params in request

a-zharkov:
In case if we need to get any customer orders, then we need to pass parameter `status=any` to this request.
Thank you.

### DIFF
--- a/resources/customer.js
+++ b/resources/customer.js
@@ -63,11 +63,12 @@ Customer.prototype.sendInvite = function sendInvite(id, params) {
  * Get all orders belonging to a customer.
  *
  * @param {Number} id Customer ID
+ * @param {Object} [params] Query parameters
  * @return {Promise} Promise that resolves with the result
  * @public
  */
-Customer.prototype.orders = function orders(id) {
-  const url = this.buildUrl(`${id}/orders`);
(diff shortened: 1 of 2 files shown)


lpinca:
This isn't documented, can you please add a test?
It's ok to just copy https://github.com/MONEI/Shopify-api-node/blob/b2a9adce3654d834e452545e937b6dfba4b8c8ed/test/customer.test.js#L149-L158

and add a dummy query parameter like it's done here https://github.com/MONEI/Shopify-api-node/blob/b2a9adce3654d834e452545e937b6dfba4b8c8ed/test/customer.test.js#L64

Thank you.

a-zharkov:
Yes, sure, will do now. Shopify don't document a lot of things :)

lpinca:
Would you mind fixing the lint issue?

a-zharkov:
I used github editor, okay let me check in IDE.

a-zharkov:
@lpinca Done. Could you please advise when the next release will be, because we need this fix in our project in the code that already on production. So asap release will be very helpful. Thank you!

lpinca:
I'll merge this and cut a new release later today.

lpinca:
Thank you.

PR #231, opened 2018-09-20
[ts] add confirmed to IOrder

squirly:
This is not explicitly documented but shows up in the documentation response object and in the wild.

See View Response at https://help.shopify.com/en/api/reference/orders/order#show


lpinca:
Thank you.

PR #245, opened 2018-12-04
Add missing recurringApplicationCharge.customize endpoint support

jcbarton:
See https://help.shopify.com/en/api/reference/billing/recurringapplicationcharge#customize

Specify the chargeId, and the new capped amount


lpinca:
Thank you.

PR #249, opened 2019-01-08
[ts] Add missing properties to Transaction interface.

msoghoian:
Added missing properties to ITransaction:
* message
* parent_id
* currency_exchange_adjustment

Added new interface `ICurrencyExchangeAdjustment` as defined here:
https://help.shopify.com/en/api/reference/orders/transaction


lpinca:
Thank you.

PR #253, opened 2019-01-21
change test property to reflect proper type

### LINKED ISSUE #252: Invalid definition for interface ICreateRecurringApplicationCharge in index.d.ts
in index.d.ts, ICreateRecurringApplicationCharge is defined as: 
```
interface ICreateRecurringApplicationCharge {
    capped_amount?: number;
    name: string;
    price: number;
    return_url: string;
    terms?: string;
    test?: true | null;
    trial_days?: number;
    trial_ends_on?: string;
  }
```

This causes an error when trying to set `test: true` because according to the interface, `test` is defined as type `true` vs `boolean`:

```
[ts]
Argument of type '{ test: boolean; name: string; price: number; return_url: string; }' is not assignable to parameter of type 'ICreateRecurringApplicationCharge'.
Types of property 'test' are incompatible.
Type 'boolean' is not assignable to type 'true'. [2345]
```

nikmartin:
fixes #252

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -573,7 +573,7 @@ declare namespace Shopify {
         price: string;
         return_url: string;
         status: "accepted" | "declined" | "expired" | "pending";
-        test: true | null;
+        test: boolean | null;
         updated_at: string;
     }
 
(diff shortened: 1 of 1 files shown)


lpinca:
Thank you.

PR #254, opened 2019-01-25
add 'investigate' to risk enum

nikmartin:
per: https://help.shopify.com/en/api/reference/orders/order-risk#properties


nikmartin:
wrong branch :( - will re-do this request

PR #255, opened 2019-01-25
add 'investigate' to order risk enum

nikmartin:
per: https://help.shopify.com/en/api/reference/orders/order-risk#properties


lpinca:
Thank you.

PR #264, opened 2019-04-08
[ts] Add support for multi-currency model changes

msoghoian:
Went through this guide for migrating to multi-currency support  https://help.shopify.com/en/api/guides/multi-currency-migration-guide

Mainly added `*_set` properties, which include the amount/currency that the shop has set, and then the conversion of what the customer has set as their preferred currency (presentment). Here's an example:
```
"amount_set": {
  "shop_money": {
    "amount": "5.00",
        "currency_code": "USD"
  },
  "presentment_money": {
    "amount": "3.96",
        "currency_code": "EUR"
  }
}
```

These updates affect the following interfaces:
* ICheckoutOrder
* ICustomer
* ILineItemTaxLine
* ILineItem
* IOrderShippingLine
* IOrder
* IProductVariant
* IRefundLineItem
* IShop
* ITransaction

I also added the interface `IDiscountApplication` onto `IOrder`. I don't know that it applies to this multi-currency ticket but it was missing.

There is also some very light alphabetization of some a few properties.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -658,6 +658,7 @@ declare namespace Shopify {
     order?: ICheckoutOrder;
     payment_url?: string;
     phone: string | null;
+    presentment_currency: string;
     referring_site: string;
     request_details?: ICheckoutRequestDetails;
     requires_shipping?: boolean;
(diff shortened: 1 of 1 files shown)


lpinca:
Thank you.

msoghoian:
@lpinca Would it be possible to tag a new release? I am in need of these changes to the model to support multi-currency in my app. Thanks!

lpinca:
Done.

ns8-matt:
@lpinca Thank you! Much appreciated!

PR #267, opened 2019-04-09
Add missing properties and alphabetization to several interfaces

msoghoian:
Another pass at maintaining this. Added missing properties to several interfaces, did a bit of alphabetization for readability.


lpinca:
Thanks.

PR #268, opened 2019-04-17
[api] Add optional params to inventoryLevels in the Location resource

sarchila:
Unfortunately, the params are undocumented in https://help.shopify.com/en/api/reference/inventory/location#inventory_levels but this resource does indeed take query params including `limit` and `page`, which are crucial if one is getting inventory levels for a location that has many product variants/inventory items.

### DIFF
--- a/resources/location.js
+++ b/resources/location.js
@@ -25,7 +25,7 @@ assign(Location.prototype, omit(base, ['create', 'delete', 'update']));
  * Retrieves a list of inventory levels for a location.
  *
  * @param {Number} id Location ID
- * @param {Object} [params] Optional query params
+ * @param {Object} [params] Query parameters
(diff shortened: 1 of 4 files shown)


resources/location.js
```diff
@@ -28,9 +28,9 @@ assign(Location.prototype, omit(base, ['create', 'delete', 'update']));
  * @return {Promise} Promise that resolves with the result
  * @public
  */
-Location.prototype.inventoryLevels = function inventoryLevels(id) {
+Location.prototype.inventoryLevels = function inventoryLevels(id, params) {
```
lpinca: Can you please add JSDoc for the `params` argument?

sarchila: Sure thing, done. 👍

lpinca:
Thank you.

PR #270, opened 2019-05-03
Add missing type definition for storefrontAccessToken

joshgare:
There was no definition for storefrontAccessToken despite there being a interface for IStorefrontToken. This PR adds this definition and allows Typescript to recognise the methods in storefrontAccessToken.


lpinca:
Thank you.

PR #274, opened 2019-05-20
Add source_identifier to Order interface

joshgare:
Order is missing the property source_identifier when using this node_module in Typescript. This PR will add the missing property to the interface.


lpinca:
Thank you.

PR #276, opened 2019-05-23
Update types

msoghoian:
* `OrderFinancialStatus` was missing `refunded`
* `OrderFulfillmentStatus` was missing `restocked`
* `IOrderAdjustmentMoney` had `amount: number`, but everything i'm seeing in the documentation as well as in the API responses, this value is a `string`. I added it as an OR because I didn't want to cause any breaking changes for anyone depending on that property being a number for whatever reason.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2197,7 +2197,7 @@ declare namespace Shopify {
   }
 
   interface IOrderAdjustmentMoney {
-    amount: number;
+    amount: number | string;
     currency_code: string;
   }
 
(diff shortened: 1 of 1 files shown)


lpinca:
Would you be so kind to prefix commit titles with "[ts]" so I can use the "Rebase and merge" button :).
Thank you.

msoghoian:
You got it! I will try to remember to always do this in the future. Do you prefer one commit per PR or do you like to have smaller more detailed commits like i've done here?

lpinca:
One logical change per commit so it's easy to revert if needed. What you've done here is good.

PR #277, opened 2019-05-24
added missing types

jineshshah36:


index.d.ts
```diff
@@ -2513,6 +2516,10 @@ declare namespace Shopify {
   interface ITRansactionReceipt {
     testcase: boolean;
     authorization: string;
+    balance_transaction?: {
```
lpinca: Is there any documentation available for this?

jineshshah36: the docs say it could have anything in there depending on the gateway so maybe it would be better to make it more flexible

jineshshah36: updated it to be:

```typescript
interface ITRansactionReceipt {
    testcase: boolean;
    authorization: string;
    balance_transaction?: { [k: string]: any };
}
```

PR #278, opened 2019-05-27
[ts] Revise interface ILineItem

Marko-Markovic-5:
price should be a string
add `not_eligible` to possible values of fulfillment status
ref: https://help.shopify.com/en/api/reference/orders/order#line-items-property


lpinca:
Thank you.

Issue #273, opened 2019-05-14
IFulfillmentService has no `id` field


MicheleLambertucci:
Looking at the Shopify API Documentation, [each fulfillment service has an `id` field](https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillmentservice#index). 

The current `IFulfillmentService` in `index.d.ts` doesn't have the `id` field.

lpinca:
PR's welcome.

PR #279, opened 2019-05-29
[ts] Add missing properties to IFulfillmentService

msoghoian:
These are partially undocumented properties that are not yet listed in the main table of properties here: https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillmentservice#properties

But these properties are returned with the api calls to this endpoint, and are documented in the response objects on that same page: https://help.shopify.com/en/api/reference/shipping-and-fulfillment/fulfillmentservice#index

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1486,13 +1486,17 @@ declare namespace Shopify {
 
   interface IFulfillmentService {
     callback_url: string;
+    email: string | null;
     format: 'json';
     handle: string;
+    id: number;
+    include_pending_stock: boolean;
     inventory_management: boolean;
     location_id: number;
     name: string;
     provider_id: number | null;
     requires_shipping_method: boolean;
+    service_name: string | null;
     tracking_support: boolean;
   }
 


index.d.ts
```diff
@@ -1486,13 +1486,17 @@ declare namespace Shopify {
...
     format: 'json';
     handle: string;
+    id: number;
+    include_pending_stock: boolean;
     inventory_management: boolean;
     location_id: number;
     name: string;
     provider_id: number | null;
     requires_shipping_method: boolean;
+    service_name: string | null;
```
lpinca: Can this be `null`?

lpinca:
It seems this also fixes #273.

index.d.ts
msoghoian: I'm not sure, but because it's officially undocumented in the table of properties, I did not want to set it as a hard requirement. 

Based on the numerous API calls i've made to shopify, I think this is probably always only going to be a `string`, but I don't know that 100% for sure. I would have no arguments if you think we should remove `null` here.

lpinca:
Thank you.

PR #282, opened 2019-06-12
feature: added support for graphql

leamarty:
Added support for GraphQL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -594,8 +594,9 @@ shopify.graphql(`{
       }
     }
   }
-}`).then(customers => console.log(customers))
-   .catch(err => console.error(err));
+}`)
+.then(customers => console.log(customers))
+.catch(err => console.error(err));
 ```
 
 ## Become a master of the Shopify ecosystem by:
--- a/README.md
+++ b/README.md
@@ -97,6 +97,14 @@ argument.
 shopify.on('callLimits', limits => console.log(limits));
 ```
 
+When using the GraphQL API, a different property is used to track the API call
+limit: `callGraphqlLimits`.
+
+```js
+shopify.on('callGraphqlLimits', limits => console.log(limits));
+```
+
+
 ### Resources
 
 Every resource is accessed via your `shopify` instance:
@@ -564,6 +572,31 @@ shopify.metafield.create({
 
 where `params` is a plain JavaScript object. See https://help.shopify.com/api/reference?ref=microapps
 for parameters details.
+
+
+### GraphQL
+
+The `shopify` instance also allows to use the GraphQL API through the `graphql` method, which returns a
+promise that resolves with the result data:
+
+```js
+const shopify = new Shopify({
+  shopName: 'your-shop-name',
+  accessToken: 'your-oauth-token'
+});
+
+shopify.graphql(`{
+  customers(first: 5) {
+    edges {
+      node {
+        displayName
+        totalSpent
+      }
+    }
+  }
+}`).then(customers => console.log(customers))
+   .catch(err => console.error(err));
+```
 
 ## Become a master of the Shopify ecosystem by:
 
--- a/README.md
+++ b/README.md
@@ -100,6 +100,8 @@ shopify.on('callLimits', limits => console.log(limits));
 When using the GraphQL API, a different property is used to track the API call
 limit: `callGraphqlLimits`.
 
+Keep in mind that the `autoLimit` option is ignored while using GraphQL API.
+
 ```js
 shopify.on('callGraphqlLimits', limits => console.log(limits));
 ```
@@ -576,8 +578,8 @@ for parameters details.
 
 ### GraphQL
 
-The `shopify` instance also allows to use the GraphQL API through the `graphql` method, which returns a
-promise that resolves with the result data:
+The `shopify` instance also allows to use the GraphQL API through the `graphql`
+method, which returns a promise that resolves with the result data:
 
 ```js
 const shopify = new Shopify({
--- a/index.js
+++ b/index.js
@@ -141,17 +141,17 @@ Shopify.prototype.request = function request(url, method, key, params) {
 /**
  * Updates GraphQL API call limits.
  *
- * @param {String} throttleStatus The throttle status returned in the GraphQL response, under "extensions"
+ * @param {String} throttle The status returned in the GraphQL response
  * @private
  */
-Shopify.prototype.updateGraphqlLimits = function updateGraphqlLimits(throttleStatus) {
-  const callGraphqlLimits = this.callGraphqlLimits;
+Shopify.prototype.updateGqlLimits = function updateGqlLimits(throttle) {
+  const limits = this.callGraphqlLimits;
 
-  callGraphqlLimits.remaining = throttleStatus.currentlyAvailable;
-  callGraphqlLimits.current = throttleStatus.maximumAvailable - throttleStatus.currentlyAvailable;
-  callGraphqlLimits.max = throttleStatus.maximumAvailable;
+  limits.remaining = throttle.currentlyAvailable;
+  limits.current = throttle.maximumAvailable - throttle.currentlyAvailable;
+  limits.max = throttle.maximumAvailable;
 
-  this.emit('callGraphqlLimits', callGraphqlLimits);
+  this.emit('callGraphqlLimits', limits);
 };
 
 /**
@@ -180,8 +180,8 @@ Shopify.prototype.graphql = function graphql(data) {
 
   return got(options).then(res => {
     const body = JSON.parse(res.body);
-    if (body.extensions && body.extensions.cost && body.extensions.cost.throttleStatus) {
-      this.updateGraphqlLimits(body.extensions.cost.throttleStatus);
+    if (body.extensions && body.extensions.cost) {
+      this.updateGqlLimits(body.extensions.cost.throttleStatus);
     }
     return body.data || {};
   });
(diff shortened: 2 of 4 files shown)


lpinca:
- Was this tested against a real shop?
- Documentation is missing
- [Rate limiting is different](https://help.shopify.com/en/api/graphql-admin-api/graphql-admin-api-rate-limits) so we should at least document that the `autoLimit`  option, the `callLimits` attribute, and the `'callLimits'` event do not work when using GraphQL.

leamarty:
@lpinca added full tests, documentation, and new `callGraphqlLimits` attribute (and event).

And yes, it has been tested against real shop.

README.md
```diff
@@ -97,6 +97,14 @@ argument.
 shopify.on('callLimits', limits => console.log(limits));
 ```
 
+When using the GraphQL API, a different property is used to track the API call
```
lpinca: We should still mention that the `autoLimit` option is ignored.

README.md
```diff
@@ -565,6 +573,32 @@ shopify.metafield.create({
 where `params` is a plain JavaScript object. See https://help.shopify.com/api/reference?ref=microapps
 for parameters details.
 
+
+### GraphQL
+
+The `shopify` instance also allows to use the GraphQL API through the `graphql` method, which returns a
```
lpinca: Please keep the limit of 80 chars per line everywhere.

index.js
```diff
@@ -133,6 +138,58 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+ */
+Shopify.prototype.graphql = function graphql(data) {
+  const url = assign({ path: '/admin/api/graphql.json' }, this.baseUrl);
+  const options = assign({
+    headers: {
+      'User-Agent': `${pkg.name}/${pkg.version}`,
+      'Content-Type': 'application/graphql'
+    },
+    timeout: this.options.timeout,
+    json: false,
```
lpinca: Does this default to `true`? If not please remove it.

index.js
```diff
@@ -133,6 +138,58 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+    body: data
+  }, url);
+
+  if (this.options.accessToken) {
+    options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
+  }
+
+  return got(options).then(res => {
+    const body = res.body ? JSON.parse(res.body) : {};
+    this.updateGraphqlLimits(body.extensions && body.extensions.cost && body.extensions.cost.throttleStatus);
```
lpinca: I think this makes the

```js
if (!throttleStatus) return;
```

check above redundant? I would just call `this.updateGraphqlLimits()` if `throttleStatus` exists.

index.js
```diff
@@ -133,6 +138,58 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+    return body.data || {};
+  }, err => {
+    return Promise.reject(err);
```
lpinca: This is not needed. You can remove the whole error handler.

test/shopify.test.js
```diff
@@ -367,4 +373,168 @@ describe('Shopify', () => {
...
+      scope
+        .post(path)
+        .reply(200, { extensions: dummyExtensions });
+
+      return shopify.graphql(dummyData)
+        .then(() => {
+          expect(shopify.callGraphqlLimits).to.deep.equal({
+            remaining: 997,
+            current: 3,
+            max: 1000.0
```
lpinca: Can this be a non integer?

index.js
```diff
@@ -133,6 +138,58 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+    method: 'POST',
+    body: data
+  }, url);
+
+  if (this.options.accessToken) {
+    options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
+  }
+
+  return got(options).then(res => {
+    const body = res.body ? JSON.parse(res.body) : {};
```
lpinca: Are there cases where the body is empty? It seems odd with GraphQL.

README.md
leamarty: Right, on it

README.md
leamarty: Ok

index.js
leamarty: True

index.js
leamarty: Right

index.js
leamarty: Done

test/shopify.test.js
leamarty: Its weird, but this is an example of a response I got from Shopify:
```
{"data":{"customers":{"edges":[{"node":{"displayName":"John Doe","totalSpent":"8800.00"}}]}},"extensions":{"cost":{"requestedQueryCost":7,"actualQueryCost":3,"throttleStatus":{"maximumAvailable":1000.0,"currentlyAvailable":997,"restoreRate":50.0}}}}
```

So, seems it can be a float

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
   });
 };
 
+/**
+ * Updates GraphQL API call limits.
+ *
+ * @param {String} throttleStatus The throttle status returned in the GraphQL response, under "extensions"
```
```suggestion
 * @param {String} throttleStatus The throttle status returned in the GraphQL
 *     response, under "extensions".
```

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+ * @private
+ */
+Shopify.prototype.updateGraphqlLimits = function updateGraphqlLimits(throttleStatus) {
```
lpinca: 80 chars per line please.

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+  const callGraphqlLimits = this.callGraphqlLimits;
+
+  callGraphqlLimits.remaining = throttleStatus.currentlyAvailable;
+  callGraphqlLimits.current = throttleStatus.maximumAvailable - throttleStatus.currentlyAvailable;
```
lpinca: Ditto, I will not add more comments like this but there are more occurrences also in the tests.

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+    body: data
+  }, url);
+
+  if (this.options.accessToken) {
+    options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
+  }
+
+  return got(options).then(res => {
+    const body = JSON.parse(res.body);
+    if (body.extensions && body.extensions.cost && body.extensions.cost.throttleStatus) {
```
lpinca: Questions: Do we have to check also for both `cost` and `cost.throttleStatus`? Are there cases where `extensions` exists and `cost` or `cost.throttleStatus` are not present?

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+  const url = assign({ path: '/admin/api/graphql.json' }, this.baseUrl);
+  const options = assign({
+    headers: {
+      'User-Agent': `${pkg.name}/${pkg.version}`,
+      'Content-Type': 'application/graphql'
+    },
+    timeout: this.options.timeout,
+    retries: 0,
+    method: 'POST',
+    body: data
```
lpinca: Is building `data` user responsibility? Can the user use variables? If so is the following signature reasonable?

```
shopify.graphql(query[, variables])
```

I think the current interface is not very appealing. I mean what is the advantage of using this module instead of got or node-fetch directly?

leamarty: I know, but this a direct request by Shopify and they want it like that, at least by now. We can think of an improvement later on, but they are rushing with this.

lpinca: Ok can you update the TypeScript definitions to include this new method?

leamarty: Done

index.js
```diff
@@ -133,6 +138,55 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+  }, url);
+
+  if (this.options.accessToken) {
+    options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
+  }
+
+  return got(options).then(res => {
+    const body = JSON.parse(res.body);
```
lpinca: If this thows, for consistency, we should throw `got.ParseError`, see

- https://github.com/sindresorhus/got/blob/v8.3.2/index.js#L380
- https://github.com/sindresorhus/got/blob/v8.3.2/errors.js#L51-L58

The options object must be built manually, we can't simply forward `options` as some options are missing.

leamarty: Done!

index.js
```diff
@@ -133,6 +139,68 @@ Shopify.prototype.request = function request(url, method, key, params) {
...
+
+  if (this.options.accessToken) {
+    options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
+  }
+
+  return got(options).then(res => {
+    try {
+      res.body = JSON.parse(res.body);
+    } catch (err) {
+      const opts = assign({
```
lpinca: This can be simplified  as follows:

```js
const opts = assign({
  host: options.hostname,
  url: urlLib.resolve(urlLib.format(options), options.path)
}, options);
```

leamarty: Right, my bad, sorry

leamarty: Done

leamarty:
Thank you! @lpinca

lpinca:
You created a mess in the history :( we have to squash before merging.

lpinca:
I will force push.

lpinca:
Ok fixed.

PR #285, opened 2019-06-13
[api] Add support for optional Shopify API version

theRealPadster:
This is just super basic, but it gets the job done. I don't have any tests for it or anything. Feel free to make changes or suggestions. I'm not sure if the request URLs always start with `/admin/`, but if they do, that part can be optimized.

### DIFF
--- a/mixins/base-child.js
+++ b/mixins/base-child.js
@@ -108,6 +108,10 @@ const baseChild = {
       .replace(/\/+/g, '/')
       .replace(/\/$/, '');
 
+    if (this.shopify.options.apiVersion) {
+      path = `/admin/api/${this.shopify.options.apiVersion}/${path.slice(7)}`;
+    }
+
     path += '.json';
 
     if (query) path += '?' + qs.stringify(query, { arrayFormat: 'brackets' });
--- a/mixins/base-child.js
+++ b/mixins/base-child.js
@@ -104,14 +104,14 @@ const baseChild = {
   buildUrl(parentId, id, query) {
     id || id === 0 || (id = '');
 
-    let path = `/admin/${this.parentName}/${parentId}/${this.name}/${id}`
-      .replace(/\/+/g, '/')
-      .replace(/\/$/, '');
+    let path = '/admin';
 
     if (this.shopify.options.apiVersion) {
-      path = `/admin/api/${this.shopify.options.apiVersion}/${path.slice(7)}`;
+      path += `/api/${this.shopify.options.apiVersion}`;
     }
 
+    path += `/${this.parentName}/${parentId}/${this.name}/${id}`;
+    path = path.replace(/\/+/g, '/').replace(/\/$/, '');
     path += '.json';
 
     if (query) path += '?' + qs.stringify(query, { arrayFormat: 'brackets' });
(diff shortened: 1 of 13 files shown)


lpinca:
This should work but I would prefer to update the `buildUrl()` function of mixins instead of doing postprocessing on the already built path.

mixins/base-child.js
```diff
@@ -108,6 +108,10 @@ const baseChild = {
       .replace(/\/+/g, '/')
       .replace(/\/$/, '');
 
+    if (this.shopify.options.apiVersion) {
+      path = `/admin/api/${this.shopify.options.apiVersion}/${path.slice(7)}`;
```
lpinca: I would prefer something like this:

```js
let path = '/admin';

if (this.shopify.options.apiVersion) {
  path += `/api/${this.shopify.options.apiVersion}`;
}

path += `/${this.parentName}/${parentId}/${this.name}/${id}`;
path = path.replace(/\/+/g, '/').replace(/\/$/, '') + '.json';
```

We should also update resources that overrides `buildUrl()`. There are a few.

theRealPadster: Alright, that's slightly cleaner. I found overrides in shopify-payments and fulfillment-events, so I've updated those as well. I am working on adding tests, but I've never used instanbul/nyc or nock before, just dabbled a bit with mocha/chai, so it's slightly slow going.

lpinca: No problem, take you time.

theRealPadster: I'm not sure how I'd check to make sure the request path changed, since it looks like the `Shopify.request` method just returns the body. Is there a way to intercept it before it goes out?

lpinca: You can use nock. Use the updated path on it. See https://github.com/nock/nock#read-this---about-interceptors.

theRealPadster: Does it actually transform the request? I tried this, but it's hitting the `/admin/test` one, not the API versioned one. 

```
scope
  .get(`/admin/api/${apiVersion}/test`)
  .reply(200, { status: 'success' });

scope
  .get('/admin/test')
  .reply(400, { status: 'fail' });

const shopify = new Shopify({ shopName, accessToken, apiVersion });
const url = assign({ path: '/admin/test' }, shopify.baseUrl);

return shopify.request(url, 'GET', 'foo')
  .then((res) => {
    expect(res).to.equal({ status: 'success' });
  });
```

lpinca: you are not using the mixins `buildUrl()` in that example are you are directly calling `shopify.request()`. Try with a resource that actually uses the mixin and it should work.

test/shopify.test.js
```diff
@@ -303,6 +304,55 @@ describe('Shopify', () => {
         });
     });
 
+    // Test base mixin
+    it('injects the api version to the path if provided (1/4)', () => {
```
lpinca: Can you move these in tests of the respective resources? For example this should be in `shop.test.js`.
We should ideally do it for all resources, but I think I'm fine doing it only for some just to cover all cases.

theRealPadster: Alright. I also removed the checks after the requests complete since it's really just testing if the proper URL gets hit. The tests were just copy/pasted from existing tests anyways. Is this fine?

lpinca: I would keep it. I understand it is redudand but it does not harm and makes the test work like a real example.

One final request: can you also please make the newly added `shopify.graphql()` support versions? Thank you.

lpinca:
Thank you.

theRealPadster:
Glad I could contribute :)

Issue #288, opened 2019-06-22
Type definition improvements.


YourWishes:
I've been working with the API and I have a few ideas for type definition improvements, I'll submit a few pull requests when I can find some time as I've made a few of these already.

### Fixes
 - ICarrierService - Needs `id:number` as part of the interface

### Improvements
 - WebhookRequestMap - Interface for lookup of WebhookTopic to an interface of webhook request. e.g. WebhookTopic 'products/create' can have request of IProduct. This will require additions to what can be given by a webhook, e.g.
   - ICart - Interface for cart webhook callbacks
   - IDeleteItem - Interface for a webhook callback for deleted item, generally `{ id:number }`
   - ICollection - See further below
 - ICollection - Interface for a standard collection model, since ISmartCollection and ICustomCollection have a fair bit of overlap, also the various collection webhooks e.g. `/collections/create` return essentially either `ISmartCollection | ICustomCollection`
 - ICarrierRequest - Interface for carrier request callbacks, will also require ICarrierItem and ICarrierAddress
 - ICarrierResponse - Interface for carrier responses to Shopify
 - Better use of params for various functions, e.g. `product.list` currently accepts `params?:any`
 - Better use of similar object types. e.g. ILineItemProperty vs ICheckoutLineItemProperty, essentially doing the same thing.

I guess there's also a case to be made for converting the project from Javascript to Typescript, if this is something that is being considered.

DaveLo:
I was going to create an issue related to Pagination, but this seems like the right place to add.

My static type checker is unhappy with `order.nextPageParameters` as the response is defined as `IOrder[]`.  I'm not super knowledgeable in typescript, but I think maybe something like:

```typescript
interface IOrderList extends Array<IOrder> {
  nextPageParameters?: string,
  previousPageParameters?: string,
}
```

Alternatively maybe a response interface that could contain any of the typed array responses so as not to need to add a list type for every endpoint with a list api call?

bags1976:
I had a similar issue with paginated results on products.  Something as simple as a generic this
```
interface IPaginatedResult<T> extends Array<T> {
		nextPageParameters?: any;
		previousPageParameters?: any,
}
```

Then change the return types for all paginated responses (not all are currently implemented according to the docs as of today).

```
order: {
    cancel: (id: number, params?: any) => Promise<Shopify.IOrder>;
    close: (id: number) => Promise<Shopify.IOrder>;
    count: (params?: any) => Promise<number>;
[...]

PR #289, opened 2019-06-23
Multiple TS Changes and Improvements

YourWishes:
See #288 

Added missing property `id` to ICarrierService
Added Interface for ICart, ICartLineItem and IDeletedItem
Added WebhookType map, e.g. `WebhookType<'products/create'>` will yeild `IProduct`
Added various types for shipping carrier callbacks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -955,6 +955,7 @@ declare namespace Shopify {
   }
 
   interface ICarrierService {
+    id:number;
     active: boolean;
     callback_url: string;
     carrier_service_type: string; // I think this could be restricted to "api" or "legacy"
--- a/index.d.ts
+++ b/index.d.ts
@@ -2618,6 +2618,21 @@ declare namespace Shopify {
     user_type: UserType;
   }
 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2617,6 +2618,21 @@ declare namespace Shopify {
     user_type: UserType;
   }
 
+  type ICartLineItem = ICheckoutLineItem;
+
+  interface ICart {
+    id:string,
```
lpinca: Replace all `,` with `;`?

index.d.ts
```diff
@@ -2617,6 +2618,21 @@ declare namespace Shopify {
...
+    token:string,
+    line_items: Shopify.ICartLineItem[],
+    note: string|null,
+    updated_at: string,
+    created_at: string
+  }
+
+  export type IDeletedItem = {
+    id:number
```
lpinca: Missing `;`?

lpinca:
Also please remove the final dot from the last commit title and run prettier on all commits. Thank you.

PR #290, opened 2019-06-26
Multiple TS Changes and Improvements

YourWishes:
Refer to https://github.com/MONEI/Shopify-api-node/pull/289
Also see #288 

Also, is there a preferred prettier config? I went with `--single-quote --trailing-comma es5`, although it's done some aesthetic damage to my WebhookMap which could be a tad confusing to glance at, thoughts?

### Pull Request Fixes
Fixed commit title to omit trailing period
Ran through prettier to correct semicolon and comma inconsistencies.

### Definition Changes
Added missing property id to ICarrierService
Added Interface for ICart, ICartLineItem and IDeletedItem
Added WebhookType map, e.g. WebhookType<'products/create'> will yeild IProduct
Added various types for shipping carrier callbacks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -955,6 +955,7 @@ declare namespace Shopify {
   }
 
   interface ICarrierService {
+    id: number;
     active: boolean;
     callback_url: string;
     carrier_service_type: string; // I think this could be restricted to "api" or "legacy"
--- a/index.d.ts
+++ b/index.d.ts
@@ -2618,6 +2618,21 @@ declare namespace Shopify {
     user_type: UserType;
   }
 
+  type ICartLineItem = ICheckoutLineItem;
+
+  interface ICart {
+    id: string;
+    token: string;
+    line_items: Shopify.ICartLineItem[];
+    note: string | null;
+    updated_at: string;
+    created_at: string;
+  }
+
+  export type IDeletedItem = {
+    id: number;
+  };
+
   export type WebhookTopic =
     | 'app/uninstalled'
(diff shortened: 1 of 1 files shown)


lpinca:
No trailing commas please. Single quote is ok and consistent with current style.

YourWishes:
> 
> 
> No trailing commas please. Single quote is ok and consistent with current style.

alright thanks for clearing up, I ran the defintions through prettier with the updated config, no changes to commit though.

lpinca:
Rubber stamp LGTM. Are there breaking changes? I mean, do these types work with TypeScript < 3? Is there a chance that this will break people build? (I don't use TypeScript)

YourWishes:
The map for the webhooks uses conditional types, [according to the docs](https://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) these were introduced in TS 2.8.x.

Reverting back to version 2.0.10 shows that it will still compile without errors when not being used, the only time an error will be thrown is when attempting to use the new maps in a format that is unexpected.

### TS 2.0.10
```
type T = Shopify.WebhookType<'products/create'>;
let x = request.get() as T;//T is type of 'products/create' not the expected IProduct
x.id;//Will result in an error since TS thinks x is a string
```

### TS 2.8.x+
```
type T = Shopify.WebhookType<'products/create'>;
let x = request.get() as T;//T is type of IProduct
x.id;//Works without error.
```

lpinca:
Thank you.

PR #291, opened 2019-07-05
Add GraphQL variables

joelvh:
This PR allows specifying GraphQL variables to the request. Keeps backward-compatibility.

### DIFF
--- a/index.js
+++ b/index.js
@@ -209,6 +209,18 @@ Shopify.prototype.graphql = function graphql(data, variables) {
     if (res.body.extensions && res.body.extensions.cost) {
       this.updateGqlLimits(res.body.extensions.cost.throttleStatus);
     }
+
+    if (res.body.errors) {
+      const first = res.body.errors[0];
+      const err = new Error(first.message);
+      err.locations = first.locations;
+      err.path = first.path;
+      err.extensions = first.extensions;
+      err.errors = res.body.errors;
+
+      throw err;
+    }
+
     return res.body.data || {};
   });
 };
--- a/index.js
+++ b/index.js
@@ -185,7 +185,6 @@ Shopify.prototype.graphql = function graphql(data, variables) {
     timeout: this.options.timeout,
     retries: 0,
     method: 'POST',
-    responseType: 'json',
     body
   }, url);
 
--- a/index.js
+++ b/index.js
@@ -185,6 +185,7 @@ Shopify.prototype.graphql = function graphql(data, variables) {
     timeout: this.options.timeout,
     retries: 0,
     method: 'POST',
+    responseType: 'json',
     body
   }, url);
 
--- a/index.js
+++ b/index.js
@@ -174,7 +174,7 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
   path += '/graphql.json';
 
-  const json = variables != null;
+  const json = variables !== undefined && variables !== null;
   const body = json ? JSON.stringify({
     query: data,
     variables
--- a/index.js
+++ b/index.js
@@ -165,7 +165,7 @@ Shopify.prototype.updateGqlLimits = function updateGqlLimits(throttle) {
  * @return {Promise}
  * @public
  */
-Shopify.prototype.graphql = function graphql(data) {
+Shopify.prototype.graphql = function graphql(data, variables) {
   let path = '/admin/api';
 
   if (this.options.apiVersion) {
@@ -174,16 +174,22 @@ Shopify.prototype.graphql = function graphql(data) {
 
   path += '/graphql.json';
 
+  const json = variables !== null;
+  const body = json ? JSON.stringify({
+    query: data,
+    variables
+  }) : data;
+
   const url = assign({ path }, this.baseUrl);
   const options = assign({
     timeout: this.options.timeout,
     retries: 0,
     method: 'POST',
-    body: data
+    body
   }, url);
 
   options.headers['User-Agent'] = `${pkg.name}/${pkg.version}`;
-  options.headers['Content-Type'] = 'application/graphql';
+  options.headers['Content-Type'] = json ? 'application/json' : 'application/graphql';
 
   if (this.options.accessToken) {
     options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
@@ -203,7 +209,6 @@ Shopify.prototype.graphql = function graphql(data) {
     if (res.body.extensions && res.body.extensions.cost) {
       this.updateGqlLimits(res.body.extensions.cost.throttleStatus);
     }
-
     return res.body.data || {};
   });
 };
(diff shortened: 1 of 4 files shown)


joelvh:
@lpinca not sure if you're the primary guy, but wanted to see if you know why the coverage test is failing?

lpinca:
probably because the additional if branch you added is not covered by a test.

alexandresaiz:
@lpinca hey luigi. cana you decide on this PR?

lpinca:
I will not review it until checks are green.

lpinca:
FWIW I'm :+1: on the feature. I suggested it when GraphQL support was added. See https://github.com/MONEI/Shopify-api-node/pull/282#discussion_r293100938.

joelvh:
Got tests passing.

joelvh:
@lpinca merge?

lpinca:
Thanks, I will take a look when a I can.

index.js
```diff
@@ -167,7 +167,7 @@ Shopify.prototype.updateGqlLimits = function updateGqlLimits(throttle) {
  * @return {Promise}
  * @public
  */
-Shopify.prototype.graphql = function graphql(data) {
+Shopify.prototype.graphql = function graphql (data, variables) {
```
```suggestion
Shopify.prototype.graphql = function graphql(data, variables) {
```

index.js
```diff
@@ -176,16 +176,23 @@ Shopify.prototype.graphql = function graphql(data) {
 
   path += '/graphql.json';
 
+  const json = typeof variables !== 'undefined' && variables !== null;
```
```suggestion
  const json = variables !== undefined && variables !== null;
```

index.js
```diff
@@ -176,16 +176,23 @@ Shopify.prototype.graphql = function graphql(data) {
...
+    variables
+  }) : data;
+
   const url = assign({ path }, this.baseUrl);
   const options = assign({
     timeout: this.options.timeout,
     retries: 0,
     method: 'POST',
-    body: data
+    responseType: 'json',
```
lpinca: I think this option does not exist on got@8.3.1 It should be `json` and set only when the `json` variable above is `true`.

index.js
```diff
@@ -176,16 +176,23 @@ Shopify.prototype.graphql = function graphql(data) {
...
+    body
   }, url);
 
   options.headers['User-Agent'] = `${pkg.name}/${pkg.version}`;
-  options.headers['Content-Type'] = 'application/graphql';
+  options.headers['Content-Type'] = json ? 'application/json' : 'application/graphql';
```
lpinca: Set this to `'application/graphql'` only when `json` is `false`. The `json` option will take care of adding `'application/json'` when it is set.

index.js
```diff
@@ -206,6 +213,17 @@ Shopify.prototype.graphql = function graphql(data) {
       this.updateGqlLimits(res.body.extensions.cost.throttleStatus);
     }
 
+    if (res.body.errors) {
+      const first = res.body.errors[0];
+      const err = new Error(first.message);
```
lpinca: With the REST API this is one of https://github.com/sindresorhus/got/tree/v8.3.1#errors but I don't any of those classes makes sense here. I think a generic `Error` is ok. What do you think?

index.js
```diff
@@ -206,6 +213,17 @@ Shopify.prototype.graphql = function graphql(data) {
...
+      err.locations = first.locations;
+      err.path = first.path;
+      err.extensions = first.extensions;
+      err.errors = res.body.errors;
```
```suggestion
      err.response = res
```

lpinca: For consistency with the REST API.

index.js
joelvh: Isn't the response always JSON?

lpinca: Yes but afaik this option is not available on got@8.3.1 so it probably has not effect here. Makes sense?

lpinca: See, https://github.com/sindresorhus/got/tree/v8.3.2 there is no `responseType` option. Anyway on second look we try to `JSON.parse()` so we don't need this option anyway or we will end parsing twice. I think we should completely remove this option.

index.js
lpinca: No longer valid

test/shopify.test.js
```diff
@@ -403,6 +403,64 @@ describe('Shopify', () => {
       });
     });
 
+    it('returns an Error with GraphQL info when the request fails', () => {
```
lpinca:
```suggestion
    it('returns an Error with GraphQL info if the request fails (1/2)', () => {
```

test/shopify.test.js
```diff
@@ -403,6 +403,64 @@ describe('Shopify', () => {
...
+      }, err => {
+        expect(err).to.be.an.instanceof(Error);
+        expect(err.message).to.equal(message);
+        expect(err.locations).to.deep.equal(locations);
+        expect(err.path).to.equal(path);
+        expect(err.extensions).to.deep.equal(extensions);
+      });
+    });
+
+    it('returns an Error with GraphQL info when the request fails with variables', () => {
```
lpinca:
```suggestion
    it('returns an Error with GraphQL info if the request fails (2/2)', () => {
```

test/shopify.test.js
```diff
@@ -540,7 +598,7 @@ describe('Shopify', () => {
         });
     });
 
-    it('returns a valid response when using graphql endpoint', () => {
+    it('returns a valid response when using graphql endpoint without variables', () => {
```
lpinca:
```suggestion
    it('returns a valid response when using graphql endpoint (1/2)', () => {
```

test/shopify.test.js
```diff
@@ -553,6 +611,19 @@ describe('Shopify', () => {
         .then(res => expect(res).to.deep.equal(response.data));
     });
 
+    it('returns a valid response when using graphql endpoint with variables', () => {
```
lpinca:
```suggestion
    it('returns a valid response when using graphql endpoint (2/2)', () => {
```

index.js
joelvh: That’s what I ended up doing. Will push the update shortly.

index.js
joelvh: Yeah, generic error is simplest for now I think.

index.js
alexandresaiz: @joelvh any way I can DM ?

joelvh: hi @alexandresaiz sent you an email.

joelvh:
@lpinca resolved conflicts - your updates should all be there.

index.js
lpinca: It seems this regressed :)

test/shopify.test.js
lpinca: It seems these have not been applied?

joelvh:
Now they should be there.

lpinca:
Thank you.

joelvh:
Yeah, something happened with the merge conflicts :-/ All good now. Thanks for approving!

agentlewis:
Just want to express my gratitude for this code - exactly what I needed! It will tidy up my string literals and also enable me to use types in my code! 🙏

PR #294, opened 2019-07-16
Add Option for Including Presentment Prices Header

jbgrunewald:
This is a quick fix to add an option while instantiating the Shopify object so the necessary header will be included to get the presentment prices.

### DIFF
--- a/index.js
+++ b/index.js
@@ -114,15 +114,10 @@ Shopify.prototype.request = function request(url, method, key, params) {
     method
   }, url);
 
+  options.headers['User-Agent'] = `${pkg.name}/${pkg.version}`;
+
   if (this.options.accessToken) {
     options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
-  }
-
-  if (this.options.presentmentPrices
-    && /(product(?!_)|variant|product_listing)[s]*/.test(key)
-    && !url.path.includes('smart_collection')
-    && method === 'GET') {
-    options.headers['X-Shopify-Api-Features'] = 'include-presentment-prices';
   }
 
   if (params) {
--- a/index.js
+++ b/index.js
@@ -22,8 +22,8 @@ const pkg = require('./package');
  * @param {String} options.password The private app password
  * @param {String} options.accessToken The persistent OAuth public app token
  * @param {String} [options.apiVersion] The Shopify API version to use
- * @param {Boolean} [options.presentmentPrices] Whether to include the header
- *   to pull presentment prices for products
+ * @param {Boolean} [options.presentmentPrices] Whether to include the header to
+ *     pull presentment prices for products
  * @param {Boolean|Object} [options.autoLimit] Limits the request rate
  * @param {Number} [options.timeout] The request timeout
  * @constructor
@@ -59,6 +59,7 @@ function Shopify(options) {
 
   this.baseUrl = {
     auth: !options.accessToken && `${options.apiKey}:${options.password}`,
+    headers: {},
     hostname: !options.shopName.endsWith('.myshopify.com')
       ? `${options.shopName}.myshopify.com`
       : options.shopName,
@@ -73,12 +74,6 @@ function Shopify(options) {
 
     this.request = stopcock(this.request, conf);
   }
-
-  this.presentmentHeader = {
-    headers: {
-      ['X-Shopify-Api-Features']: 'include-presentment-prices'
-    }
-  };
 }
 
 Object.setPrototypeOf(Shopify.prototype, EventEmitter.prototype);
@@ -114,7 +109,6 @@ Shopify.prototype.updateLimits = function updateLimits(header) {
  */
 Shopify.prototype.request = function request(url, method, key, params) {
   const options = assign({
-    headers: {},
     timeout: this.options.timeout,
     json: true,
     retries: 0,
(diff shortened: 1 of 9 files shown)


index.js
```diff
@@ -117,6 +118,10 @@ Shopify.prototype.request = function request(url, method, key, params) {
     options.headers['X-Shopify-Access-Token'] = this.options.accessToken;
   }
 
+  if (this.options.presentmentPrices) {
+    options.headers['X-Shopify-Api-Features'] = 'include-presentment-prices';
```
lpinca: I would prefer this header to be only set for the product and productVariant resources.

jbgrunewald: Done

As a side note, I think the conditional logic is rather bloated. If you have any clout with the Shopify folks, we might try to persuade them to make this an option in the individual requests. Requiring this as a header is outside the norm for their API's and annoying to handle consistently.

lpinca: My idea was to actually add the header in the implementation of the relevant resources. For example overriding `buildUrl()` in https://github.com/MONEI/Shopify-api-node/blob/master/resources/product.js and adding the header if the option is set.

We can then set the `User-Agent` and `X-Shopify-Access-Token` headers after the options object is built. Does it make sense?

jbgrunewald: I was thinking something similar, but was looking at the functions in the base-child mixin. I like your proposal but it's unclear to me how the overriding works in this scenario. If I create a BuildUrl field on the product prototype, won't it be overwritten with the BuildUrl in base when we call assign(Product.Prototype, base)?

The other concern was with visibility about the headers being set. Everything for the requests is being handled in that one function right now. If we start setting headers inside the resources, it won't be immediately obvious. This could get confusing and potentially lead to bugs later if someone adds a default header to the option object due to a change in the APIs. That default header will get overridden when these resources are used.

lpinca: Yes I recognise that current design does not allow to do this in a clean way but I don't like this regex based approach. If there is no other way I would prefer to go back, revert eeb6bcb, and set it on every request.

jbgrunewald: I've addressed the feedback by overloading the base functions where required in the resources. I have concerns about the headers being shadowed later with how the assign function works, but I suppose unit testing will have to serve as a defense for now.

index.js
```diff
@@ -22,6 +22,7 @@ const pkg = require('./package');
  * @param {String} options.password The private app password
  * @param {String} options.accessToken The persistent OAuth public app token
  * @param {String} [options.apiVersion] The Shopify API version to use
+ * @param {Boolean} [options.presentmentPrices] Whether to include the header to pull presentment prices for products
```
```suggestion
 * @param {Boolean} [options.presentmentPrices] Whether to include the header to
       pull presentment prices for products
```

index.js
```diff
@@ -113,6 +114,8 @@ Shopify.prototype.request = function request(url, method, key, params) {
     method
   }, url);
 
+  options.headers['User-Agent'] = `${pkg.name}/${pkg.version}`;
```
lpinca: Can you please use an empty object on line 110?

resources/product-listing.js
```diff
@@ -19,7 +19,38 @@ function ProductListing(shopify) {
   this.key = 'product_listing';
 }
 
-assign(ProductListing.prototype, omit(base, ['create', 'update']));
+assign(ProductListing.prototype, omit(base, ['create', 'update', 'buildUrl', 'productIds', 'count', 'delete']));
```
lpinca: Use `pick` instead to assign only `get`,  and `list`.

resources/product-listing.js
```diff
@@ -19,7 +19,38 @@ function ProductListing(shopify) {
...
+ * @param {Number|String} [id] Record ID
+ * @param {Object} [query] Query parameters
+ * @return {Object} URL object
+ * @private
+ */
+ProductListing.prototype.buildUrl = function(id, query) {
+  const baseUrl = base.buildUrl.call(this, id, query);
+
+  if (this.shopify.options.presentmentPrices) {
+    baseUrl.headers = { ['X-Shopify-Api-Features']: 'include-presentment-prices' };
```
lpinca: Style nit here and everywhere else: please try to keep the limit of 80 chars per line.

test/common.js
```diff
@@ -11,22 +11,33 @@ const apiKey = 'bc731e500840231da5b43bb3f388d2f0';
 const apiVersion = '2019-04';
 const shopName ='johns-apparel';
 
-const shopify = new Shopify({ shopName, accessToken });
+const shopify = new Shopify({ shopName, accessToken, presentmentPrices:true });
```
lpinca:
```suggestion
const shopify = new Shopify({ shopName, accessToken, presentmentPrices: true });
```

resources/product-listing.js
```diff
@@ -19,7 +19,38 @@ function ProductListing(shopify) {
...
+
+/**
+ * Builds the request URL.
+ *
+ * @param {Number|String} [id] Record ID
+ * @param {Object} [query] Query parameters
+ * @return {Object} URL object
+ * @private
+ */
+ProductListing.prototype.buildUrl = function(id, query) {
```
lpinca: I know this follows my suggestion but on second thought I think it's easier to override only the methods that need the special header instead of the `buildUrl()`. It should keep things a little cleaner.

lpinca:
Thank you.

Issue #88, opened 2016-03-18
errors with webpack


Jan-Jan:
Sorry to bother: 

I get two errors when using nodify-shopify with webpack

```
ERROR in ./~/nodify-shopify/index.js
Module not found: Error: Cannot resolve module 'fs' in /Users/naturgewalt/Coding/gigster/vavavoo-slingshot/node_modules/nodify-shopify
 @ ./~/nodify-shopify/index.js 6:11-24
```

```
ERROR in ./~/nodify-shopify/package.json
Module parse failed: ./node_modules/nodify-shopify/package.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "name": "nodify-shopify",
|   "version": "0.2.1",
|   "description": "Shopify API client",
 @ ./~/nodify-shopify/index.js 8:12-32

ERROR in ./~/nodify-shopify/~/got/package.json
Module parse failed: ./node_modules/nodify-shopify/node_modules/got/package.json Line 2: Unexpected token :
You may need an appropriate loader to handle this file type.
| {
|   "name": "got",
|   "version": "6.2.0",
|   "description": "Simplified HTTP/HTTPS requests",
 @ ./~/nodify-shopify/~/got/index.js 21:12-37
```

I can make the latter go away by adding to webpack.conf
```
loaders: [{ test: /\.json$/, loader: "json-loader" }]
```
but no other package requires this...

The primary I can make the parse error go away using
```
target: 'node'
```
but, then I get a failure
```
Uncaught ReferenceError: require is not defined
```
(which is understandable since I not compiling for the web)

lpinca:
This is a bit problematic.

The first issue is [`fs.readdirSync`](https://github.com/microapps/Nodify-Shopify/blob/dc7cdc7f4c30d532902aa3119aeea6c7492eadf7/index.js#L79). I'm not familiar with webpack but with browserify you can use the `brfs` transform and get the list of resources as expected in the bundle. I guess the same can be achieved in webpack.

The second issue is that resources are [required and instantiated lazily](https://github.com/microapps/Nodify-Shopify/blob/dc7cdc7f4c30d532902aa3119aeea6c7492eadf7/index.js#L80-L93). This will not work with bundlers because AFAIK they can only read `require()` calls that are statically analyzable.

So I think that right now there is no way to use the lib in the browser unless we remove the lazy require feature.

lpinca:
[...]

PR #296, opened 2019-08-02
Lazily register resources without reading the file system

joelvh:
By removing the need to read from the file system, all dependencies can be packaged into one file with things like Webpack.

### DIFF
--- a/resources/index.js
+++ b/resources/index.js
@@ -62,6 +62,9 @@ const RESOURCES = {
   webhook: 'webhook'
 };
 
+//
+// Require and instantiate the resources lazily.
+//
 function registerAll(Shopify) {
   Object.keys(RESOURCES).forEach(prop => {
     Object.defineProperty(Shopify.prototype, prop, {
--- a/resources/index.js
+++ b/resources/index.js
@@ -1,0 +1,84 @@
+const RESOURCES = {
+  accessScope: 'access-scope',
+  apiPermission: 'api-permission',
+  applicationCharge: 'application-charge',
+  applicationCredit: 'application-credit',
+  article: 'article',
+  asset: 'asset',
+  balance: 'balance',
+  blog: 'blog',
+  carrierService: 'carrier-service',
+  checkout: 'checkout',
+  collect: 'collect',
+  collectionListing: 'collection-listing',
+  comment: 'comment',
+  country: 'country',
+  currency: 'currency',
+  customCollection: 'custom-collection',
+  customerAddress: 'customer-address',
+  customerSavedSearch: 'customer-saved-search',
+  customer: 'customer',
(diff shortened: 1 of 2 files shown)


lpinca:
See https://github.com/MONEI/Shopify-api-node/issues/88.

joelvh:
I’ve worked around it until now, this update just removes the file system dependency. Acceptable?

lpinca:
I think it complicates things as we have to update the map every time we add a new resource.

joelvh:
I thought about that, but how often is that? You have to add the resource file at that point, as well as  tests. So it would be natural to add a line to the resources as a 3rd place to be updated.

lpinca:
Counter point, how often is this module bundled and why would you do it?

joelvh:
Serverless. All our integrations are going the serverless route, in this case using Lambda functions. There’s a 50mb file size limit. But regardless, the way to package and upload is to use Webpack. We have many services that leverage this package.

I had thought about creating a way to eagerly load each resource as well, so we can package only what we need. That would be a second phase, assuming this is going to be accepted. (Basically, adding the ability to import each resource I want and “register” individually.)

lpinca:
Let me think about it (I'm not a fan of the resources map, I would rather remove the lazy load functionality).

joelvh:
Agreed. Without lazy loading, we could have a way to require resources individually, but would still need to add new require statements for new resources.

lpinca:
Thank you.

joelvh:
Thanks @lpinca ! Would you accept a pull request with a refactoring to make lazy loading optional?

lpinca:
If it turns out to be small and clean I'm fine with it. Let's also keep open the option of completely removing the lazy loading feature (it would probably be a breaking change).

resources/index.js
```diff
@@ -0,0 +1,87 @@
...
+  Object.keys(RESOURCES).forEach(prop => {
+    Object.defineProperty(Shopify.prototype, prop, {
+      get: function get() {
+        const resource = require(`./${RESOURCES[prop]}`);
+
+        return Object.defineProperty(this, prop, {
+          value: new resource(this)
+        })[prop];
+      },
+      set: function set(value) {
```
joelvh: @lpinca reviewing this, does the `value` ever get used after being set since the getter always instantiates a value?

lpinca: Yes, see https://github.com/MONEI/Shopify-api-node/blob/master/test/shopify.test.js#L89-L99.

lpinca: After being set is will be on the object, not the prototype so the get is no longer used.

joelvh: Ok makes sense. Thanks.

PR #298, opened 2019-08-08
Add polling for 202 response

### LINKED ISSUE #261: Retrieving Shipping Rates
The shipping rates API has a special case where it returns a 202 status code, this means the shipping rates are not available yet and should be retried later. Right now I don't see how to distinguish between the 202 status code and 200 status code, both of which return empty arrays but need to be handled differently by the sales channel app.

Any ideas on the best way to go about this in the context of this lib? Can't think of an approach that doesn't require changing the api signature for basically everything method. Alternative is something completely separate (and hacky) for this use case.

Thoughts?

tim-phillips:
Should we stop polling after a certain number of requests? 

In which part of the readme should this be documented?

Let me know if you need any changes.

Resolves #261

### DIFF
--- a/index.js
+++ b/index.js
@@ -218,6 +218,14 @@ Shopify.prototype.graphql = function graphql(data) {
 
 resources.registerAll(Shopify);
 
-const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
+/**
+ * Waits before resolving
+ *
+ * @param {Number} ms Amount of milliseconds to wait
+ * @private
+ */
+function delay(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
 
 module.exports = Shopify;
--- a/index.js
+++ b/index.js
@@ -131,6 +131,14 @@ Shopify.prototype.request = function request(url, method, key, params) {
 
     this.updateLimits(res.headers['x-shopify-shop-api-call-limit']);
 
+    if (res.statusCode === 202) {
+      const retryAfter = res.headers['retry-after'] * 1000 || 0;
+      const path = urlLib.parse(res.headers['location']).path;
+      const newUrl = assign({ path }, this.baseUrl);
+      return delay(retryAfter)
+        .then(() => this.request(newUrl, 'GET', key));
+    }
+
     if (key) return body[key];
     return body || {};
   }, err => {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -210,4 +218,6 @@ Shopify.prototype.graphql = function graphql(data) {
 
 resources.registerAll(Shopify);
 
+const delay = ms => new Promise(resolve => setTimeout(resolve, ms));
```
lpinca: Can you please use a function declaration and add JSDoc?

test/checkout.test.js
```diff
@@ -69,7 +69,24 @@ describe('Shopify#checkout', () => {
...
+  it('completes a free checkout and retries 202 response', () => {
+    const token = 'b490a9220cd14d7344024f4874f640a6';
+    const baseUrl = shopify.baseUrl;
+    const pathname = `/admin/checkouts/${token}/complete.json`;
+    const href = `${baseUrl.protocol}//${baseUrl.hostname}${pathname}`;
+    const output = fixtures.res.complete;
+
+    scope
+      .post(pathname, {})
+      .reply(202, output, { Location: href, 'Retry-After': 0 })
```
lpinca:
```suggestion
      .reply(202, '', { Location: href, 'Retry-After': 0 })
```

Shouldn't body be empty?

test/checkout.test.js
```diff
@@ -110,4 +127,21 @@ describe('Shopify#checkout', () => {
...
+  it('gets a list of shipping rates and retries 202 response', () => {
+    const token = 'exuw7apwoycchjuwtiqg8nytfhphr62a';
+    const baseUrl = shopify.baseUrl;
+    const pathname = `/admin/checkouts/${token}/shipping_rates.json`;
+    const href = `${baseUrl.protocol}//${baseUrl.hostname}${pathname}`;
+    const output = fixtures.res.shippingRates;
+
+    scope
+      .get(pathname)
+      .reply(202, [], { Location: href, 'Retry-After': 0 })
```
lpinca:
```suggestion
      .reply(202, '', { Location: href, 'Retry-After': 0 })
```

lpinca:
> Should we stop polling after a certain number of requests?

I don't know, I think we can assume that Shopify will not send 202 forever.

> In which part of the readme should this be documented?

None I would treat it as a bug fix.

lpinca:
Thank you.

PR #312, opened 2019-10-31
add interface params to IInventoryItem

builtbylane:
adds 'country_code_of_origin', 'harmonized_system_code', and 'requires_shipping' to `IInventoryItem`


lpinca:
Thank you.

PR #316, opened 2019-11-10
Added productResourceFeedback

pliew:
Sales channel specific command.  ResourceFeedback is on the application level whereas this is on the product level

### DIFF
--- a/resources/product-resource-feedback.js
+++ b/resources/product-resource-feedback.js
@@ -16,42 +16,13 @@ function ProductResourceFeedback(shopify) {
   this.shopify = shopify;
 
   this.parentName = 'products';
-  this.name = 'resource_feedback';
-  this.key = 'resource_feedback';
+  this.key = this.name= 'resource_feedback';
 }
 
 assign(ProductResourceFeedback.prototype, pick(baseChild, [
   'buildUrl',
-  'count',
-  'delete'
+  'create',
+  'list'
 ]));
-
-/**
- * Creates a new product variant.
- *
- * @param {Number} productId Product ID
- * @param {Object} params Product variant properties
- * @return {Promise} Promise that resolves with the result
- * @public
- */
-ProductResourceFeedback.prototype.create = function(productId, params) {
-  const url = this.buildUrl(productId);
-
-  return this.shopify.request(url, 'POST', this.key, params);
-};
-
-/**
- * Get a list of product variants.
- *
- * @param {Number} productId Product ID
- * @param {Object} [params] Query parameters
- * @return {Promise} Promise that resolves with the result
- * @public
- */
-ProductResourceFeedback.prototype.list = function(productId, params) {
-  const url = this.buildUrl(productId, undefined, params);
-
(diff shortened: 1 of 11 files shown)


lpinca:
Can you please change the name from `productResourceFeedback` to `resourceFeedback` as per https://help.shopify.com/en/api/reference/sales-channels/resourcefeedback?

lpinca:
Perhaps something like this https://github.com/MONEI/Shopify-api-node#metafields using `resource_type` and `resource_id` can be used with the already existing `ResourceFeedback` to work on a specific resource?

pliew:
> ~Can you please change the name from `productResourceFeedback` to `resourceFeedback` as per https://help.shopify.com/en/api/reference/sales-channels/resourcefeedback?~
> 
> Is there any official documentation for this? I can't find it.

https://help.shopify.com/en/api/reference/sales-channels/productresourcefeedback
Looks like they hide it but it's been available since 2019.  We've been discussing this with the internal team at Shopify and it's not a new feature

pliew:
> Perhaps something like this https://github.com/MONEI/Shopify-api-node#metafields using `resource_type` and `resource_id` can be used with the already existing `ResourceFeedback` to work on a specific resource?

The reasoning I chose this method is that the resource id is based off the REST url and not based on a post/query parameter.

lpinca:
As long as it is documented I'm fine with it. Give me some time to review. Thanks.

resources/product-resource-feedback.js
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Creates a new product variant.
+ *
+ * @param {Number} productId Product ID
+ * @param {Object} params Product variant properties
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+ProductResourceFeedback.prototype.create = function(productId, params) {
```
lpinca: The default one from the base-child mixing should work unless am I missing something can you please remove?

resources/product-resource-feedback.js
```diff
@@ -0,0 +1,57 @@
...
+ * @param {Shopify} shopify Reference to the Shopify instance
+ * @constructor
+ * @public
+ */
+function ProductResourceFeedback(shopify) {
+  this.shopify = shopify;
+
+  this.parentName = 'products';
+  this.name = 'resource_feedback';
+  this.key = 'resource_feedback';
```
lpinca:
```suggestion
  this.key = this.name = 'resource_feedback';
```

resources/product-resource-feedback.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param {Shopify} shopify Reference to the Shopify instance
+ * @constructor
+ * @public
+ */
+function ProductResourceFeedback(shopify) {
+  this.shopify = shopify;
+
+  this.parentName = 'products';
+  this.name = 'resource_feedback';
```
lpinca:
```suggestion
```

lpinca:
Can you please also add documentation for this on README.md?

PR #319, opened 2019-11-13
add `tax_code: string;` to `IProductVariant`

builtbylane:
update index.d.ts: add `tax_code: string;` to `IProductVariant`. When not present, shopify returns an empty string `''`


lpinca:
Thank you.

PR #321, opened 2019-11-15
typings: tax_code should be nullable

squirly:
Minor error in the TypeScript definition:

In GraphQL this field is not required https://help.shopify.com/en/api/graphql-admin-api/reference/object/productvariant#taxcode-2019-10
and the JSON API docs suggest this as well:
> This parameter applies only to the stores that have the Avalara AvaTax app installed

~ https://help.shopify.com/en/api/reference/products/product-variant#tax-code-property-2019-10

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2142,7 +2142,7 @@ declare namespace Shopify {
     requires_shipping: boolean;
     sku: string;
     taxable: boolean;
-    tax_code: string;
+    tax_code: string | null;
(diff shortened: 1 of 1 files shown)


lpinca:
Hmm, it was added in https://github.com/MONEI/Shopify-api-node/pull/319 and author said that when not present Shopify returns an empty string.

lpinca:
Thank you....

PR #322, opened 2019-11-16
Update customer type for IDraftOrder

baorv:


index.d.ts
```diff
@@ -1384,14 +1384,34 @@ declare namespace Shopify {
...
+  }
 
   interface IDraftOrder {
     applied_discount: IDraftOrderDiscount[];
     billing_address: ICustomerAddress;
     completed_at: string | null;
     created_at: string;
     currency: string;
-    customer: string;
+    customer: IDraftOrderCustomer;
```
lpinca: I think we can use  the already existing `ICustomer` interface.

baorv: I updated code to use ICustomer interface.

lpinca:
Thank you.

PR #328, opened 2019-12-27
Adding provinces to the file definition of IShippingZoneCountry

brennick:
Currently the API returns provinces but the type definitions for IShippingZoneCountry does not contain it. This PR fixes that issue.


lpinca:
Thank you.

PR #340, opened 2020-01-29
Fulfillment orders

Stumblor:
Added in the api endpoint Fulfilment Orders from the new 2020-01 api, and added unit tests.

Hopefully I got everything right! Let me know if you need any changes made, cheers.


Stumblor:
Let me review the above - also I think some of the existing endpoints are having trouble with the new api version (webhooks for instance)

Stumblor:
Prettier is asking for changes that would make things different to how things are currently - could you advise the best thing to do please?

lpinca:
@Stumblor Thank you for the PR. I've opened #341 instead of adding a lot of comments here. I hope you don't mind.

If you can address the TODO task listed on #341 by opening a new PR against the `add/fulfillment-order-resource` branch it would be greatly appreciated.

PR #341, opened 2020-02-02
Add FulfillmentOrder resource

lpinca:
TODO:

- [x] Add TypeScript type definitions.


Stumblor:
Apologies @lpinca, but I'm not seeing any of the previous typescript defs?

PR #344, opened 2020-02-17
Pr/1

grepx:


lpinca:
The proposed changes cannot be applied. See https://github.com/MONEI/Shopify-api-node/blob/932ca09b05b416e391b060809cd4b94174c3d339/index.d.ts#L2090-L2097

PR #363, opened 2020-03-21
ICheckout's applied discount property is an object

matomesc:
Although not documented, I can confirm from ~50k checkouts that `ICheckout.applied_discount` is a `ICheckoutDiscount` object not an array.


lpinca:
Thank you.

PR #366, opened 2020-04-01
Added check for graphql extensions.cost.throttleStatus

mete89:
Shopify doesn't return throttleStatus for "staff" plan which causes the error below.

TypeError: Cannot read property 'currentlyAvailable' of undefined
    at Shopify.updateGraphqlLimits (/var/app/current/node_modules/shopify-api-node/index.js:191:31)
    at /var/app/current/node_modules/shopify-api-node/index.js:234:12
    at tryCatcher (/var/app/current/node_modules/bluebird/js/release/util.js:16:23)

### DIFF
--- a/index.js
+++ b/index.js
@@ -229,7 +229,7 @@ Shopify.prototype.graphql = function graphql(data, variables) {
   }
 
   return got(uri, options).then((res) => {
-    if (res.body.extensions && res.body.extensions.cost) {
+    if (res.body.extensions && res.body.extensions.cost && res.body.extensions.cost.throttleStatus) {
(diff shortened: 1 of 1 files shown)


lpinca:
Can you do something like this

https://github.com/MONEI/Shopify-api-node/blob/3af24fa80204f7faf09d3392075b293da08ccad4/index.js#L92

instead for consistency with `Shopify.prototype.updateLimits()`?

Thanks.

mete89:
Hi @lpinca,

Does it look good now?

lpinca:
Yes but please fix the lint error.

lpinca:
Thank you.

PR #368, opened 2020-04-01
[ts] Fixes to Order line item and tax line typings (#367)

### LINKED ISSUE #367: Incorrect line item type definitions
There are some mistakes in the line item type definitions.
For refund line items:
* `product_id` is marked as non-null
* `requires_shipping` is incorrectly named as `require_shipping`
* `tip_payment_gateway` and `tip_payment_method` are marked as always being present

For order line items:
* `discount_allocations` is missing
* `fulfillment_status` is given the values of order fulfillment states rather than line item ones
* `tip_payment_gateway` and `tip_payment_method` are missing
* `variant_id` is marked as nullable

tkalliom:
Fixes #367. The typings have had separate types `ILineItem`, for the `line_item` property of refund line items, and `IOrderLineItem`, for the `line_items` property of orders. Both typings have some issues, as detailed in the bug report. As the refund `line_item` is specified in the API documentation using a reference to order line items, this commit removes the former type in favor of the latter to reflect the specification.

Also fixes order tax lines being typed with the `price_set` property, which only exists on line item level tax lines.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1795,34 +1795,6 @@ declare namespace Shopify {
     price_set: IOrderAdjustmentAmountSet;
   }
 
-  interface ILineItem {
-    discount_allocations: ILineItemDiscountAllocation[];
-    fulfillable_quantity: number;
-    fulfillment_service: string;
-    fulfillment_status: LineItemFulfillmentStatus;
-    gift_card: boolean;
-    grams: number;
-    id: number;
-    name: string;
-    price: string;
-    price_set: IOrderAdjustmentAmountSet;
-    product_id: number;
-    properties: ILineItemProperty[];
-    quantity: number;
-    require_shipping: boolean;
-    sku: string;
-    tax_lines: ILineItemTaxLine[];
-    taxable: boolean;
-    tip_payment_gateway: string;
-    tip_payment_method: string;
-    title: string;
-    total_discount: string;
-    total_discount_set: IOrderAdjustmentAmountSet;
-    variant_id: number;
-    variant_title: string;
-    vendor: string;
-  }
-
   interface ILocation {
     id: number;
     active: boolean;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2038,7 +2010,6 @@ declare namespace Shopify {
   interface IOrderTaxLine {
     title: string;
     price: string;
-    price_set: IOrderAdjustmentAmountSet;
```
lpinca, line 2041: Added in https://github.com/MONEI/Shopify-api-node/commit/f71a10a3b588d1e889d792188dcc498d1afa301e I don't think it should be removed.

index.d.ts
```diff
@@ -2060,14 +2032,16 @@ declare namespace Shopify {
     requires_shipping: boolean;
     sku: string;
     title: string;
-    variant_id: number | null;
```
lpinca, line 2063: `null` was added in https://github.com/MONEI/Shopify-api-node/commit/b3b34ccf4f92e8292096c8660793c47922599f83 I think it should not be removed.

index.d.ts
tkalliom: At that time, this `IOrderTaxLine` type was used for both order-level `tax_lines` (which don’t have `price_set`) and item-level `tax_lines` (which do have `price_set`). This change reorganizes types so that this is only used for order-level ones, so this should not have `price_set`.

index.d.ts
tkalliom: I relied on the documentation too much. It’s not mentioned at https://shopify.dev/docs/admin-api/rest/reference/orders/order#line-items-property-2020-01 but according to https://community.shopify.com/c/Shopify-APIs-SDKs/Orders-line-item-variant-id-fields-are-null/td-p/130401 yes, this can be `null`

lpinca:
Thank you.

Issue #324, opened 2019-11-19
autoLimit should actually check Shopify response and wait to retry


fancydev18:
Hi,

We're looking at switching to this library instead of https://github.com/christophergregory/shopify-node-api

There's an issue with autoLimit. The other library accepted a limit and it automatically checks Shopify's response with the remaining limit, and waits until it retries again. Simply said, it just works, we never had to worry about limits. Some calls will take longer to return.

Here we tried 
```
autoLimit: { calls: 1, interval: 1000, bucketSize: 16 }, 
```
according to https://github.com/MONEI/Shopify-api-node/issues/135

and it still gives error
```
"statusCode": 429,
	"statusMessage": "Too Many Requests",
```
after ~10 seconds. 

We can see the Shopify limit decreasing with ```.on('callLimits')```, but the library seems to ignore it.
Could it check that and wait automatically?

Checking it would also help if having multiple instances making simultaneous calls. Since the current limit is always returned by Shopify, no need to worry what limits to set in each case.

Thanks

lpinca:
The limit is per instance, if you have multiple instances you need another strategy. See discussion in https://github.com/MONEI/Shopify-api-node/issues/199.

fancydev18:
Right, it turns out there was another instance on the same app. This is quite complex to manage.

The fix seems simple: could you please check for 429 code, wait a little and retry? Here's how the other library does it in 4 lines:
https://github.com/christophergregory/shopify-node-api/blob/master/lib/shopify.js#L196

It would all be hidden from this library users and it will work no matter how many instances are used (across multiple servers).

lpinca:
No, that's wrong, you should never get 429, it's a user error. Tweak the `autoLimit` config to handle multiple instances or use a queue. I will not add a retry + backoff solution based on the 429 status code.

If you want it disable the `autoLimit` option and implement it yourself, the `http.ServerResponse` and its status code is attached to the `error` object.

```js
p.catch((err) => console.log(err.response.statusCode))
```

joantune:
@lpinca  - 'just use a queue' - erm we have a hint from the server on both the response code and the headers.
Why is it `wrong`? it's definitely something I'm willing to implement - as it's much easier than the answer of: change your whole architecture
Serverless? tough luck - make it a queue somehow.. that somehow takes into account the time..

joantune:
Also - the Retry-after header exists for the same purpose - from Shopify's docs:
[...]

Issue #199, opened 2018-07-15
Add backoff and retry option


corbinu:
Hello,

The autoLimit works great when doing a bunch of operations in a row. However I have noticed in the last few days that Shopify must have speedup something in their system as I am getting Product/Update webhooks called against my API at faster than their own API limit. Each product update hook makes a single call for that products metafields so the autoLimit doesn't help. Thoughts?

lpinca:
I'm not sure I understand the issue, are webhooks triggered too frequently? In that case the only thing to do is to reduce the request rate and the `autoLimit` option already allows to do that in a configurable way.

corbinu:
@lpinca Yes the webhooks have been triggering faster than 2 a second. The autoLimit isn't helping because there is no way for it to cross process boundaries. It would work fine if we had one process on one server but we do not. 

So Shopify seems to be calling us more like 6 a second which is getting load balanced across our servers. We then reply to each at the same rate even with the autoLimit set only 1 response per second. 

I will try to find time to open a PR tomorrow with exponential back off.

corbinu:
Just as a note we have known this was an issue and were planning to move all our webhook handlers to faktory however the business version with rate limting keeps getting delayed.

lpinca:
> The autoLimit isn't helping because there is no way for it to cross process boundaries.

Yes the state is not shared between the processes/services. That case can be addressed by allowing each process to send up to only a fraction of the total limit no? For example if you have 4 processes let each process send up to `limit/4`.

I can't see how a retry + exponential back off strategy would help for this.

corbinu:
@lpinca Yes but that both would be less performant most of the time and assumes that the number is static don't really want to have to change the value constantly around black friday for example.

Sorry I should be more clear this is happening in bursts. For example we have multiple regular and Shopify Plus clients that do large batch product updates from their ERP every few hours or every day. So ERP would be pushing like 40K product updates at once right at the API limit.

Used to be this was rare but in the last couple days it has happend everytime that happens I get emails from Shopify of our webhook failing. However I checked the logs and it is failing because we are getting these webhooks much faster than Shopify's API limits. The key is this is a burst that then dissipates. I believe that the sudden change has been caused by Shopify making a performance update on their end causing the webhooks to fire off faster. Eventually it settles down as Shopify keeps retrying and all the hooks eventually clear.

[...]

PR #377, opened 2020-05-14
Resolves #376 Retry on rate limit failure

### LINKED ISSUE #376: No Retries for 429 response on First Request from Newly Create Client
We are running into a scenario where we create a client, but we've previously hit the rate limit for that store. The newly created client doesn't have an information about our previous limits, and issues the first request. Since we've hit the request limits, we receive the 429 response. Currently, this is returned as an exception. I propose that after we update the limits in the exception handler, we recursively call the request function again and automatically retry that request.

jbgrunewald:


lpinca:
Thanks for the PR but I don't like this. It is a patch for a very specific use case (#376). Why not doing this is your code when handling the error?

Refs: https://github.com/MONEI/Shopify-api-node/issues/324

jbgrunewald:
The library claims to offer an autoLimit option for rate limiting. However it misses the scenario I’ve outlined above. The PR is specific to the case when someone wants to use the rate limiting that’s built in. If they chose not to, they can of course handle this outside of the client library. This addresses an edge case for an existing feature. 

The scenario is not really that specific to our use case. With how the client is instantiated, as a client specific to a store, it’s entirely likely that someone who has already reached the limit in another client will experience a 429 failure in the first request. Since we have nothing to capture this and update the headers, the rate limiting is t doing what it promises to do. 

James Grunewald

lpinca:
> We are running into a scenario where we create a client, but we've previously hit the rate limit for that store.

The `autoLimit` was not designed and should not handle this. If the limit is hit you should get 429, period.

Wrt to multiple clients, the `autoLimit` option was designed to work per process/instance, so if you have multiple processes/instances you have to configure it to send up to only a fraction of the total limit. For example if you have 4 processes let each process send up to limit/4.

jbgrunewald:
I’m not debating that there’s an opportunity to handle this outside if the client. What is the actual problem with this change in the client though? What harm does it cause, to automatically retry 429 responses when you expect the requests to be rate limited, but the client doesn’t have context yet? 

You said you don’t like it, but give no substantive explanation why. Everything in this client can be handled outside of the client. This change doesn’t prevent someone from getting events for hitting rate limits or controlling the rate limiter from outside of the client. It simply improves the built in rate limiting. 

James Grunewald

lpinca:
> You said you don’t like it, but give no substantive explanation why.

Because if you hit the limit you should get the 429 error.

jbgrunewald:
I’d also like to understand this reasoning in contrast to the 202 retry that is built into the client. These could also be handled outside of the client by sending a proper response object. 

James Grunewald

lpinca:
429 is a user error. 202 is not.

lpinca:
Retrying on 429 would also mask potential bugs in the `autoLimit` option and algorithm. If you just retry how do you know if you are actually sending the correct amount of requests per seconds?

jbgrunewald:
I agree, if I don’t set an auto rate limiter I should get the 429. Otherwise I should be able to automatically rate limit and the 429 is specifically an error related to rate limiting. 

Under different circumstances we should be able to set the retry option for the got client. You e mentioned how that interferes with the rate limiting in another issue. This allows for a basic retry related to the rate limiting. I don’t think it’s reasonable to expect  to get a 429 response on the first  request when I’m trying to rate limit. Without the retry there’s no way for that client to get the context though. 


James Grunewald

jbgrunewald:
You can still capture the events for the rate limiting if you are debugging. 

James Grunewald

lpinca:
> I agree, if I don’t set an auto rate limiter I should get the 429

You should also get it is the auto rate limiter is on as it means it is not working or is being used incorrectly.

jbgrunewald:
So we add a retry limit of one if that’s the concern? 

I argue that the rate limiting is a matter of timing. You can tell by listening for events what the rate limiter is doing. Assuming you needed to debug that issue. 

James Grunewald

lpinca:
It is probably not the case for Shopify but some API ban the client if 429 is returned multiple times in a small time frame. The rate limiter should only prevent 429 errors and nothing else.

jbgrunewald:
We check for the retry after header and set a delay. The rate limit headers are also received after the first failure and set on the rate limiter. Once the rate limiter has the context, you won’t be running into 429s over and over again. It does mitigate the need to try to manage the rate limiters across a distributed system that might scale dynamically though. 

Your assumption is that you always know exactly how many clients are in operation. 

James Grunewald

lpinca:
> Your assumption is that you always know exactly how many clients are in operation.

Yes, if you don't know that you can use a job queue as discussed in the other issues (#324, #199).

Again, in my opinion the whole point of using a rate limiter on the client is to never create 429 errors.

Retrying on 429 errors is a different strategy/philosophy. I want the client to never produce 429 erros.

jbgrunewald:
Again, in my opinion the whole point of using a rate limiter on the client is to never create 429 errors.

How do address creating a client and hitting the rate limit on the first request then? You need to make a request to know the existing rates.

lpinca:
> How do address creating a client and hitting the rate limit on the first request then? You need to make a request to know the existing rates.

When this happens it means there is another client running or that was running short before no? That client already knows the limits.

jbgrunewald:
> > Your assumption is that you always know exactly how many clients are in operation.
> 
> Yes, if you don't know that you can use a job queue as discussed in the other issues (#324, #199).

The job queue is the same underlying issue, you need to maintain state somewhere. If you operate an otherwise stateless distributed system, then this doesn't make sense. 

The Shopify documentation doesn't recommend doing anything related to a queue.
https://shopify.dev/concepts/about-apis/rate-limits#avoiding-rate-limit-errors

Everything mentioned there is being honored. We have caching already. I would like to add responsible retries. As mentioned, we use the headers and delay the next request. That is exactly what Shopify recommends doing. Retrying on exceptions in a production system is a common concept.

jbgrunewald:
> > How do address creating a client and hitting the rate limit on the first request then? You need to make a request to know the existing rates.
> 
> When this happens it means there is another client running or that was running short before no? That client already knows the limits.

Sure, if all of the requests are going to the same host that instantiated the original client. Assume, I have four machines that process these requests. Your solution suggests I would cache a client for each store on each machine potentially and set the rate limit to be divided evenly across each machine. 

What happens when one machine is under utilized, because the requests for that store never reach that machine? Now we are under utilizing our limits. What happens when I have 1000 stores, or 10000 stores?

lpinca:
> The job queue is the same underlying issue, you need to maintain state somewhere.

I don't understand? There are a lot job queues with built-in rate limiting for jobs. You don't need to maintain state anywhere.

> The Shopify documentation doesn't recommend doing anything related to a queue.

The Shopify documentation was changed. When this library was written they were recommending apps to never exceed the limit (https://github.com/MONEI/Shopify-api-node/issues/199#issuecomment-420056807).

lpinca:
I don't like the retrying approach but if that is the what consumer of this lib wants and what Shopify recommends I'm open to using it but the `autoLimit` option must be removed in this case. It no longer makes sense.

jbgrunewald:
I don't need to maintain state anywhere? At a minimum I need to manage a queue in a database. Your suggestion in the other post is to use https://github.com/OptimalBits/bull, this requires a redis store. 

So rather than responsibly use retries, I now have to manage another point of potential failure in our system and take a dependency on a library to implement the distributed queuing.  

It's obviously much easier to override the request function in this scenario and add the retry logic there. I don't agree with the opinion that a 429 should never happen. A 429 is a signal, and it should be handled appropriately. Just like you might retry for network failures or a 500. If I'm getting back other 400 level messages, then those probably can't be corrected programmatically. 

> The autoLimit option must be removed in this case. It no longer makes sense.

I don't see why that's necessary. The rate limiting is still useful, because you may want to queue requests up still. The only issue with the rate limiter, is there's no visibility into the actual available limits until after you've made at least one request. My review could be update to explicitly check for the boolean setting in the autoLimit option, which would further imply the user is not trying to control the rate limiting themselves.

lpinca:
We have different opinions. In many cases you need a job queue anyway depending on the complexity of the app.

The retrying approach with the `autoLimit` option as currently is does not make any sense to me.

jbgrunewald:
> We have different opinions. In many cases you need a job queue anyway depending on the complexity of the app.

The job in this case isn't complex, but the work is divided up by product. So for a brand with a large number of products, we make separate requests one at a time for each product. However, each request can make a variable number of requests related to the information we want. If the number of requests were always consistent, I would just rate limit as we iterated through the products. 

The opinion that 429's should never happen is extreme, and it's clear others have argued for the exact logic I'm trying to add. I agree that efforts should be made to mitigate it, and that's why the rate limiting is still useful.

> The retrying approach with the `autoLimit` option as currently is does not make any sense to me.

If we can't address hitting the rate limit initially, then I agree we should gut this from the client. Its value is minimal, and the effort to properly track and manage load to avoid 429's in a distributed system still exists. This is only useful in a small number of low scale use cases then. 

What I'm proposing is that we keep it, and we make a reasonable effort to rate limit given the context. At startup we have no context, once we do it makes no sense to be hitting those limits typically. However, it's not always possible to know what other clients might be running all the time, and also contributing to the rate limits. So in some cases it's possible to hit the 429 limits again during run time. 

We could have a different option altogether, and retry based on the header response from Shopify without any effort to rate limit the requests. However, this would mean that each new request hits the limits, then delays a retry. I'd prefer we kept the rate limiting, so when we hit the limit, we retry and other requests are then queued based on the information we have about the rate limits. So if that retry fails, we aren't just firing off other requests for no reason, because we know we hit the rate limit.

lpinca:
> The opinion that 429's should never happen is extreme

Again, it's the whole point of having a rate limiter on a client...

If you want to move this forward please remove the `autoLimit` option or change it so that when enabled it works by retrying the request. Otherwise please fork the library and change it to fit your needs.

jbgrunewald:
> Again, it's the whole point of having a rate limiter on a client...
> 
> If you want to move this forward please remove the `autoLimit` option or change it so that when enabled it works by retrying the request. Otherwise please fork the library and change it to fit your needs.

This seems like a very t"hrow the baby out with the bathwater" sort of approach. I agree that if someone sets a specific rate limit object, we should ignore the retry logic. That's a simple change to the conditional logic. We'll change the comparison to be specific to a boolean value. I don't see any reason why the rate limiter needs to be removed completely, or why it's mutually exclusive from retries. After a failure, you may want to rate limit that request and other requests with the information you have. The 429 failure is an indictor that you need to throttle, but the fact that you've hit that limit doesn't mean you need to throw away all efforts to rate limit. 

A rate limiter is a tool that can be used in a lot of ways with varying context. You seem to have a very dogmatic opinion about this. If you have some point of reference, or documentation about why rate limiting can't be used with retries that's informing these opinions, I'd be happy to take a look and educate myself further.

lpinca:
> I agree that if someone sets a specific rate limit object, we should ignore the retry logic.

It doesn't make any difference with a value set to `true`. It's still a very specific limit object when the `true` value is changed to a configuration.

I've tried to explained why in the previous comments. If the rate limiter allows you to get 429 errors, then it's broken and there is no need for it. A broken rate limiter with retries is twice as broken as it puts already delayed requests at the end of the limiter queue.

jbgrunewald:
> It doesn't make any difference with a value set to `true`. It's still a very specific limit object when the `true` value is changed to a configuration.
> 
> I've tried to explained why in the previous comments. If the rate limiter allows you to get 429 errors, then it's broken and there is no need for it. A broken rate limiter with retries is twice as broken as it puts already delayed requests at the end of the limiter queue.

You can put the configuration object into another field, rather than reusing the option field and overriding with the defaults. Also not a difficult change to make. 

Since there's no explicit guarantee about the order of the requests, I think it's fine to put them at the back of the queue. Later, an argument could be made for replacing the queue with one that only removed a request after it was successful. In that scenario, the request would continue pending until it was successful from a retry, in which case another request would be processed. 

There are numerous additions that could be made to the client to improve its flexibility and its resilience. It's clear that you aren't interesting in working toward a solution, and are likely opposed to these ideas because it conflicts with how you've designed your systems around the client. 

I suppose I'll consider your option of simply forking the library. In the meantime, here are a few resources that discuss the validity and value of using rate limiting with retries. The resource from Google specifically mentions retrying after a 429 response (and they talk about coping with retries from many clients). Based on this and the other Issues opened in the past, it's clear that popular opinion is in contrast with your own.

https://cloud.google.com/solutions/rate-limiting-strategies-techniques
https://engineering.grab.com/beyond-retries-part-1

lpinca:
FWIW I can't see where those resources suggest to use a token/leaky bucket algorithm to limit the number of requests on the client and also retry those requests when they are answered with 429 because it's a contradiction. If that happens the algorithm is broken.

> There are numerous additions that could be made to the client to improve its flexibility and its resilience. It's clear that you aren't interesting in working toward a solution, and are likely opposed to these ideas because it conflicts with how you've designed your systems around the client.

This is not true. I was open to use only a 429 retry strategy as per https://github.com/MONEI/Shopify-api-node/pull/377#issuecomment-629380602.

Also I'm more than happy to accept patches that improve the library (when they make sense).

jbgrunewald:
@lpinca You're right, there's not an article stating precisely this scenario. Ultimately, Shopify implements the leaky bucket algorithm. What the client does is operate a queue that controls its rate based on feedback from Shopify. Since it's just a queue, there's no reason why requests can't be queued again. In an ideal algorithm, the request wouldn't be removed from the queue until it completed successfully. So a request would be attempted, and could be retried any number of times for any number of reasons, based on configuration. 

You keep mentioning how the algorithm is wrong if you retry on a 429 response. It's not wrong, it's just a different solution. You're stuck on the idea that the client is implementing the leaky bucket algorithm. If it was operating a true leaky bucket, then requests wouldn't be queued and there'd be no reason to check the header responses from Shopify. There's no benefit to strictly enforcing the rate algorithm twice, because Shopify already does that by dropping the request. The client just needs to handle this gracefully. The rate limiting queue that you have, which receives feedback from Shopify, is a valuable tool combined with retries. It allows for queueing requests, and maximizes the throughput allowed by Shopify. However, there's no reason why a retry is wrong when it's used to rate limit a queue. 

https://en.wikipedia.org/wiki/Leaky_bucket#As_a_queue

Issue #378, opened 2020-05-15
Error: Missing or Invalid Options


tomgs:
Looked at https://github.com/MONEI/Shopify-api-node/issues/220, but that is not exactly my case.

I've got a public app (in draft currently), and I'm trying to call the `Shopify-api-node` module like so:

```js
const Shopify = require('shopify-api-node');
  const shopify = new Shopify({
    shopOrigin: `${shopOrigin}`,
    accessToken: `${accessToken}`
  })
```

Where both `shopOrigin` and `accessToken` are properly defined and hold the correct (to my understanding) values. More specifically:
1. My `shopOrigin` is of the form `shop-name.myshopify.com`, 
2. My`accessToken` is grabbed from the cookies provided by the [`koa-shopify-auth`](https://github.com/Shopify/quilt/tree/master/packages/koa-shopify-auth) packag

Still getting `Error: Missing or invalid options` despite this being a public app, and the fact that I only provided the `accessToken` and `shopName` parameters (and not the other, private app options).

Any thoughts?

And of course, thank you for this package! I'm attempting to integrate it with the official Node.js + React + Next tutorial, and specifically use it to skip writing proper pagination logic for the REST admin API.

**Side note:** Looking at https://github.com/MONEI/Shopify-api-node/blob/master/index.js#L31-L38, it's probably smart to add better error handling here.

Added PR https://github.com/MONEI/Shopify-api-node/pull/379, not really tested against anything for now - just let me know if it makes sense and I'll add tests, TypeScript type definitions, etc.

tomgs:
Maybe one more comment - I'm grabbing `accessToken` from the cookies provided by the [`koa-shopify-auth`](https://github.com/Shopify/quilt/tree/master/packages/koa-shopify-auth) package.

lpinca:
The `shopOrigin` option does not exist. You probably want to use the `shopName` option instead.

tomgs:
@lpinca reading is an important skill for a programmer, that's for sure. :)
Made a PR out of it, so I guess it's a net win, though.

PR #379, opened 2020-05-15
Adding clearer Shopify object instantiation error messages

tomgs:
Because I'm stupid and misread the property name in https://github.com/MONEI/Shopify-api-node/issues/378, adding error handling so it doesn't happen to anyone else.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,13 +28,36 @@ const resources = require('./resources');
  */
 function Shopify(options) {
   if (!(this instanceof Shopify)) return new Shopify(options);
-  if (
-    !options ||
-    !options.shopName ||
-    (!options.accessToken && (!options.apiKey || !options.password)) ||
-    (options.accessToken && (options.apiKey || options.password))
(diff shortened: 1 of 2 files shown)


tomgs:
Two notes:

1. No full coverage as `coveralls` mentioned above.
2. `npm run lint -- --fix` does not work, which will fail lint on the build.

lpinca:
Do we really need this level of details for 2/3 required options :(?

tomgs:
If it blocked me, it might block other people as well. Also, it's the entrypoint to the module - why not make it smoother?

tomgs:
@lpinca Should I close and just use my fork instead?

lpinca:
I would prefer to keep it as is. In my opinion this is too verbose and not really needed for just 3 options.

tomgs:
Fair enough - hope I could offer some value elsewhere in the library.
And again, thanks for writing this!

PR #389, opened 2020-07-02
Adding optional owner_resource and owner_id to the ICreateObjectMetafield. 

ckeyes88:
Shopify allows you to pass in both `owner_id` and `owner_resource` when creating a metafield ([docs here](https://shopify.dev/docs/admin-api/rest/reference/metafield?api[version]=2020-04)). This allows you to attach your metafield to a specific resource like a product or order. The current typings didn't allow those additional parameters unless I did some typescript casting so I've updating the typings here.

Happy to discuss different approaches and make any changes if you don't think this is the best solution!

- Chip

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -936,6 +936,8 @@ declare namespace Shopify {
     value: string | number;
     value_type: 'string' | 'integer';
     description?: string | null;
+    owner_id?: number;
+    owner_resource?: string;
   }
 
   interface IAsset {
--- a/test/fixtures/metafield/req/create.json
+++ b/test/fixtures/metafield/req/create.json
@@ -3,6 +3,8 @@
     "namespace": "inventory",
     "key": "warehouse",
     "value": 25,
-    "value_type": "integer"
+    "value_type": "integer",
+    "owner_id": 905684977,
+    "owner_resource": "product"
   }
 }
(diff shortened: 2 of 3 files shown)


lpinca:
Current typings allow anything see

https://github.com/MONEI/Shopify-api-node/blob/132964226ff153c97005225d6352ef13a776687c/index.d.ts#L409

and https://github.com/MONEI/Shopify-api-node#metafields.

There is no support for `/admin/orders/#{id}/metafields.json`, `/admin/blogs/#{id}/metafields.json`, `/admin/pages/#{id}/metafields.json`, etc.

lpinca:
In the context of a specific resource `owner_id` and `owner_resource` don't make much sense, see the [Create an article with a metafield](https://shopify.dev/docs/admin-api/rest/reference/online-store/article?api[version]=2020-04#create-2020-04) example.

ckeyes88:
Ah good point @lpinca. I guess my issues came from using the `ICreateObjectMetafield` type to type safe my params for that function. Looking at the source code now I see that that type was intended only for adding a metafield to another object which I agree doesn't make sense to have `owner_id` and `owner_resource`. 

Might not hurt to have some documentation there by the type to indicate that it's not intended to be used for the params of `metafield.create`.

lpinca:
> Might not hurt to have some documentation there by the type to indicate that it's not intended to be used for the params of `metafield.create`.

Isn't this already noticeable by looking at where and how the `ICreateObjectMetafield` interface is used?

lpinca:
I'm closing this. Discussion can continue if needed.

PR #392, opened 2020-07-09
[ts] Fixes to order lineitem discount fields and money type names

tkalliom:
- `{ amount, currency_code }` and `{ shop_money, presentment_money }` objects are used in many fields of the API, so having `OrderAdjustment` in the type name is misleading
- `order.line_items[*].discount_allocations[*]` have the field `amount_set`, missing from the typings


lpinca:
Thank you.

Issue #390, opened 2020-07-05
discountCode.lookup() returns 401


bastienbot:
It looks like discountCode lookup method returns a redirection to the actual resource. 
Since the url we get forwarded to is not authenticated, we get a 401, which make sense.

This crash is quite easy to reproduce, simple look up any existing code. If you look for a non existing code,  the request will return the following response without any redirection : 
```
{
"errors": "Not Found"
}
```

lpinca:
It seems that the new version of `got` (the underlying HTTP client we use in this library) does not re-add the headers when following redirects. I'm not sure if it is by design or a regression.

Can you try with shopify-api-node@2.25.1? It should work.

cc: @szmarczak

bastienbot:
I'll try with v2.25.1 this week and make the PR if I find some time in my schedule. Mean while, I have implemented a quickfix for my own code so I am ok.
Thanks for the quick follow up.

szmarczak:
@lpinca What do you mean?

```js
const got = require('got');
const http = require('http');

const server = http.createServer((request, response) => {
    if (request.url === '/redirect') {
        response.statusCode = 302;
        response.setHeader('location', '/');
        response.end();
        return;
    }
    
    response.end(JSON.stringify(request.headers));
});

server.listen(8080, async () => {
    const body = await got('http://localhost:8080', {
        headers: {
            foo: 'bar'
        }
    }).json();
    
    console.log(body); //=> {..., foo: "bar", ...}
});
```

lpinca:
Hmm, weird. it's a 303 here but I get a 401 when it is followed with `got@11`. It works with `got@8`.

szmarczak:
@lpinca The `authorization` header is cleared when there's a redirect to another hostname:

https://github.com/sindresorhus/got/blob/779062a1b99a9a1c13737b9fd613e185d97b158b/source/core/index.ts#L1193-L1211

szmarczak:
It's a safety measure. If you don't like this behavior, you need to do something like this:

```js
got('https://example.com', {
	hooks: {
		beforeRedirect: [
			(options, response) => {
				options.headers.authorization = response.request.options.headers.authorization;
			}
		]
	}
});
```

lpinca:
@szmarczak  it's not that. It seems to be a problem with URL username and password. I think they were retained even after a redirect in `got@8`.

@bastienbot are you using a private app? This patch seems to fix the issue

```diff
$ git diff
[...]

PR #396, opened 2020-07-21
[fix] Add redirect hook for authorized redirect when using apKey/pass…

kedud:
When using apiKey/password authentication in a private app, if the Shopify API returns a 303  (for instance when using `discountCode.lookup()`) then `got` follows the redirection without the apiKey & password. This behavior makes `discountCode.lookup()` return a 401 error.

To avoid the unauthorized request, a `beforeRedirect` hook is added to the `got` options.
When a redirect is followed, the API credentials are added to the followed href.

### DIFF
--- a/index.js
+++ b/index.js
@@ -119,6 +119,13 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
     timeout: this.options.timeout,
     responseType: 'json',
     retry: 0,
+    hooks: {
+      beforeRedirect: [
+        (options) => {
+          options.url.href = `${options.url.protocol}//${this.options.apiKey}:${this.options.password}@${options.url.hostname}${options.url.pathname}`;
+        }
+      ]
+    },
     method
   };
 


index.js
```diff
@@ -119,6 +119,13 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
     timeout: this.options.timeout,
     responseType: 'json',
     retry: 0,
+    hooks: {
+      beforeRedirect: [
+        (options) => {
+          options.url.href = `${options.url.protocol}//${this.options.apiKey}:${this.options.password}@${options.url.hostname}${options.url.pathname}`;
```
lpinca, line 125: What happens if the `Shopify` instance is created using the `accessToken` option instead of the `apiKey` and `password` options?

lpinca:
Thanks but I prefer this to be fixed in got. See https://github.com/MONEI/Shopify-api-node/issues/390 and https://github.com/sindresorhus/got/issues/1351.

PR #401, opened 2020-08-03
[fix] Add missing price rule typings

MitchPierias:
The `IPriceRule` interface was missing properties returned by the current [Shopify Admin API price rule request](https://shopify.dev/docs/admin-api/rest/reference/discounts/pricerule?api[version]=2020-07#)

- Added missing typings for `IPriceRule` to match current API response data
- Added test for `Shopify.prototype.updateGraphqlLimits()` and undefined arguments to meet coverage

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2202,7 +2202,20 @@ declare namespace Shopify {
   }
 
   interface IPriceRulePrerequisiteSubtotalRange {
-    prerequisite_subtotal_range: string;
+    greater_than_or_equal_to: string;
+  }
+
+  interface IPriceRulePrerequisiteQuantityRange {
+    greater_than_or_equal_to: string;
+  }
+
+  interface IPriceRulePrerequisiteToEntitlementQuantityRange {
+    prerequisite_quantity: number | null;
+    entitled_quantity: number | null;
+  }
+
+  interface IPriceRulePrerequisiteShippingPriceRange {
+    less_than_or_equal_to: string;
   }
 
   type PriceRuleTargetType = 'line_item' | 'shipping_line';
@@ -2212,7 +2225,6 @@ declare namespace Shopify {
   type PriceRuleCustomerSelection = 'all' | 'prerequisite';
 
   interface IPriceRule {
-    created_at: string;
     id: number;
     title: string;
     target_type: PriceRuleTargetType;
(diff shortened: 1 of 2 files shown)


test/shopify.test.js
```diff
@@ -532,6 +532,12 @@ describe('Shopify', () => {
       );
     });
 
+    it('returns undefined when throttle is undefined', () => {
```
lpinca, line 535: I would prefer something like this https://github.com/MONEI/Shopify-api-node/blob/3.4.2/test/shopify.test.js#L712-L723 where `res.body.extensions.cost` is an empty object.

lpinca:
Can you please force push and use only 2 commits? One for the type definitions and one for the test? Also please follow the existing commit title style. Here are two examples:

- `[ts] Add order lineitem discount amount_set to types`
- `[test] Use a smaller delay in request timeout tests`

Thank you.

test/shopify.test.js
MitchPierias: This is testing the missing like `189` coverage, when no argument is given for `throttle`, the function escapes early.
```
Shopify.prototype.updateGraphqlLimits = function updateGraphqlLimits(throttle) {
  if (!throttle) return;

  const limits = this.callGraphqlLimits;

  limits.current = throttle.maximumAvailable - throttle.currentlyAvailable;
  limits.remaining = throttle.currentlyAvailable;
  limits.restoreRate = throttle.restoreRate;
  limits.max = throttle.maximumAvailable;

  this.emit('callGraphqlLimits', limits);
};
```
Writing the test as
```
it('returns undefined when throttle is undefined', () => {
  const shopify = new Shopify({ shopName, accessToken, timeout: 100 });

  return shopify.updateGraphqlLimits().then(() => {
     expect(shopify.callGraphqlLimits).to.deep.equal({});
  });
});
```
therefore fails because we can't call `then()` on `undefined`

lpinca: Write it as follow:

```js
it('does not update callGraphqlLimits if throttleStatus is missing', () => {
  scope.post('/admin/api/graphql.json').reply(200, {
    extensions: {
      cost: {}
    }
  });

  return shopify.graphql('query').then(() => {
    expect(shopify.callGraphqlLimits).to.deep.equal({
      restoreRate: 50.0,
      remaining: 997,
      current: 3,
      max: 1000.0
    });
  });
});
```

MitchPierias: Done 👍

index.d.ts
```diff
@@ -2222,16 +2230,26 @@ declare namespace Shopify {
     value: string;
     once_per_customer: boolean;
     usage_limit: number | null;
+    allocation_limit: number | null;
     customer_selection: PriceRuleCustomerSelection;
     prerequisite_saved_search_ids: number[];
     prerequisite_subtotal_range: IPriceRulePrerequisiteSubtotalRange | null;
-    prerequisite_shipping_price_range: string;
+    prerequisite_shipping_price_range: string | null;
```
lpinca, line 2237: According to the docs this is not a string but an object like

```
{ less_than_or_equal_to: '10.0' }
```

when not `null`.

MitchPierias: Good spot! I didn't notice this one.

I've added an interface for `IPriceRulePrerequisiteShippingPriceRange`
```
interface IPriceRulePrerequisiteShippingPriceRange {
  less_than_or_equal_to: string;
}
```
and assigned it to the return type of `prerequisite_shipping_price_range`
```
interface IPriceRule {
  ...
  prerequisite_shipping_price_range: IPriceRulePrerequisiteShippingPriceRange | null;
  ...
}
```

lpinca:
Thank you.

PR #411, opened 2020-09-02
[fix] Add missing accept header

MitchPierias:
Some endpoints, such as DiscountCode lookup, require the `Accept=application/json` header to be present. The current implementation does not appear to provide the Accept header, returning a `404` when requests are made.

- Added `Accept:'application/json'` to `baseHeaders` so that it's mapped into all requests by default.
- Updated tests to check for the presence of this `Accept` header

### DIFF
--- a/index.js
+++ b/index.js
@@ -61,7 +61,10 @@ function Shopify(options) {
     protocol: 'https:'
   };
 
-  this.baseHeaders = { 'User-Agent': `${pkg.name}/${pkg.version}` };
+  this.baseHeaders = {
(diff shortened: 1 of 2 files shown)


lpinca:
This is already done by `got`, see https://github.com/sindresorhus/got/blob/v11.6.0/source/as-promise/normalize-arguments.ts#L91-L93

lpinca:
It is also tested, see https://github.com/MONEI/Shopify-api-node/blob/3.4.4/test/common.js#L25

PR #419, opened 2020-10-13
[api] Add createV2 action to Fulfillment resource

### LINKED ISSUE #418: Add support for new and recommended way of creating new fulfillments (`POST /admin/api/2020-10/fulfillments.json` without specifying an order id)
Right now, this library only supports new fulfillment creation via `shopify.fulfillment.create(orderId, params)`, which calls `POST /admin/api/2020-10/orders/{order_id}/fulfillments.json`.

It needs to add `shopify.fulfillment.create(params)` (without specifying order id) and then call `POST /admin/api/2020-10/fulfillments.json`. This is the new and recommended way of creating new fulfillments.

Docs: https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillment#createV2-2020-10

felipebrahm:
This PR adds method `shopify.fulfillment.createV2(params)` (without specifying order id) which calls `POST /admin/api/[api-version]/fulfillments.json`. This is the new and recommended way of creating new fulfillments without specifying an order id.

I decided to call it `createV2` because that's how Shopify calls the equivalent graphql mutation.

More info: https://shopify.dev/tutorials/manage-fulfillments-with-fulfillment-and-fulfillmentorder-resources
REST API Docs: https://shopify.dev/docs/admin-api/rest/reference/shipping-and-fulfillment/fulfillment#createV2-2020-10
Graphql Docs: https://shopify.dev/docs/admin-api/graphql/reference/mutation/fulfillmentcreatev2

Fixes #418

### DIFF
--- a/resources/fulfillment.js
+++ b/resources/fulfillment.js
@@ -3,6 +3,7 @@
 const assign = require('lodash/assign');
 const omit = require('lodash/omit');
 
+const base = require('../mixins/base');
 const baseChild = require('../mixins/base-child');
 
 /**
@@ -65,6 +66,19 @@ Fulfillment.prototype.cancel = function cancel(orderId, id) {
   return this.shopify
     .request(url, 'POST', undefined, {})
     .then((body) => body[this.key]);
+};
+
+/**
+ * Creates a fulfillment for one or many fulfillment orders.
+ * The fulfillment orders are associated with the same order and are assigned to the same location.
+ *
+ * @param {Object} [params] Query parameters
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+Fulfillment.prototype.createV2 = function createV2(params) {
+  const url = base.buildUrl.call(this);
+  return this.shopify.request(url, 'POST', this.key, params);
 };
 
 module.exports = Fulfillment;
--- a/resources/fulfillment.js
+++ b/resources/fulfillment.js
@@ -70,7 +70,8 @@ Fulfillment.prototype.cancel = function cancel(orderId, id) {
 
 /**
  * Creates a fulfillment for one or many fulfillment orders.
- * The fulfillment orders are associated with the same order and are assigned to the same location.
+ * The fulfillment orders are associated with the same order and are assigned to
+ * the same location.
  *
  * @param {Object} [params] Query parameters
  * @return {Promise} Promise that resolves with the result
--- a/test/fulfillment.test.js
+++ b/test/fulfillment.test.js
@@ -81,7 +81,7 @@ describe('Shopify#fulfillment', () => {
       .then((data) => expect(data).to.deep.equal(output.fulfillment));
   });
 
-  it('create a fulfillment for an order', () => {
+  it('creates a fulfillment for an order', () => {
     const input = fixtures.req.create;
     const output = fixtures.res.create;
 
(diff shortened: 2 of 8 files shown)


resources/fulfillment.js
```diff
@@ -67,4 +68,17 @@ Fulfillment.prototype.cancel = function cancel(orderId, id) {
     .then((body) => body[this.key]);
 };
 
+/**
+ * Creates a fulfillment for one or many fulfillment orders.
+ * The fulfillment orders are associated with the same order and are assigned to the same location.
```
lpinca, line 73:
```suggestion
 * The fulfillment orders are associated with the same order and are assigned to
 * the same location.
```

test/fulfillment.test.js
```diff
@@ -94,6 +94,17 @@ describe('Shopify#fulfillment', () => {
       .then((data) => expect(data).to.deep.equal(output.fulfillment));
   });
 
+  it('create a fulfillment for one or many fulfillment orders', () => {
```
lpinca, line 97:
```suggestion
  it('creates a fulfillment for one or many fulfillment orders', () => {
```

lpinca:
Can you please also update the the TypeScript type definitions?

Also, why do you say it's the recommended way? From my understanding it's only a different API not meant to replace `/admin/api/orders/{order_id}/fulfillments.json`.

Can we find a better name? I don't like `createV2`. How about `make` or  `generate` or another synonym?

lpinca:
I also wonder if this should live on the `FulfillmentOrder` resource instead of `Fulfillment`.

felipebrahm:
> Also, why do you say it's the recommended way? From my understanding it's only a different API not meant to replace `/admin/api/orders/{order_id}/fulfillments.json`.

I was quoting the article that I linked in the pull request description ([Manage fulfillments with Fulfillment and FulfillmentOrder resources](https://shopify.dev/tutorials/manage-fulfillments-with-fulfillment-and-fulfillmentorder-resources)).

Screen Shot 2020-10-15 at 4 40 28 PM


> Can we find a better name? I don't like `createV2`. How about `make` or `generate` or another synonym?

The first option I thought about was to simply use `create` for both endpoints, like this:

```javascript
Fulfillment.prototype.create = function create(parentId, params) {
  if (typeof parentId === 'object') {
    params = parentId;
    const url = base.buildUrl.call(this);
    return this.shopify.request(url, 'POST', this.key, params);
  } else {
    const url = this.buildUrl(parentId);
    return this.shopify.request(url, 'POST', this.key, params);
  }
};
```

That way you could invoke the function with either `shopify.fulfillment.create(orderId, params)` or `shopify.fulfillment.create(params)`, but I don't love this approach either because the attributes for the `params` object are completely different so it could be confusing.

I think `make` and `generate` are just synonyms of `create`, so calling it that might just cause confusion. I named it `createV2` because that's how Shopify calls this function (mutation) in their GraphQL API, so it would be easier to understand which underlying API endpoint it was calling. I agree with you that `createV2` might not be the _best_ name, but I couldn't really think of anything _better_. Thoughts?

lpinca:
Ok, thank you the explanation. Can you please update https://github.com/MONEI/Shopify-api-node/blob/3.4.4/index.d.ts with the TypeScript type definitions?

PR #420, opened 2020-10-20
[ts] Fix variant / checkout line item compare_at_price: may be null

tkalliom:
Currently the `compare_at_price` field of product variants and checkout line items are typed as required strings. However, when there is no “compare at” price, the value `null` will be present in the field.


lpinca (approved):

lpinca:
Thank you.

Issue #425, opened 2020-10-26
IInventoryItem interface is missing properties.


MBogdan18:
According to both the Shopify API documentation and the screenshot below, when I retrieve an inventory item I should be prompted with the cost field. However, I can not access it because the property doesn't exist on the interface itself and TypeScript returns an error as a result.

GET /admin/api/2020-10/inventory_items/808950810.json
docs: https://shopify.dev/docs/admin-api/rest/reference/inventory/inventoryitem#show-2020-10

Mentioning that I do have the latest version installed (3.5.1 as of the date this was written).

![Screenshot_1](https://user-images.githubusercontent.com/43849364/97166739-660f3680-178e-11eb-89b7-96f0e7ced51c.png)

![Screenshot_2](https://user-images.githubusercontent.com/43849364/97166794-7fb07e00-178e-11eb-9b30-b6546558b658.png)

lpinca:
Feel free to open a PR to fix the type definitions. It looks like `country_harmonized_system_codes` and `province_code_of_origin ` are also missing and `country_code_of_origin`, and `harmonized_system_code` can be `null`.

PR #426, opened 2020-10-27
[fix] fix inventory item interface types;

MBogdan18:
This should fix some existing properties types from the Inventory Item interface, such as `country_code_of_origin` and `harmonized_system_code` possibly being null and add `cost`, `province_code_of_origin` and `country_harmonized_system_codes`.

It's linked to this issue: https://github.com/MONEI/Shopify-api-node/issues/425

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1766,6 +1766,11 @@ declare namespace Shopify {
     processed_at: string;
   }
 
+  interface IInventoryItemCountryHarmonizedSystemCode {
+    harmonized_system_code: string | null;
+    country_code: string | null;
+  }
+
   interface IInventoryItem {
     id: number;
     sku: string;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -1766,15 +1766,23 @@ declare namespace Shopify {
     processed_at: string;
   }
 
+  interface IInventoryItemCountryHarmonizedSystemCode {
+    harmonized_system_code: string | null;
+    country_code: string | null;
```
lpinca, line 1771:
```suggestion
    harmonized_system_code: string;
    country_code: string;
```

I think these can't be `null`

index.d.ts
```diff
@@ -1766,15 +1766,23 @@ declare namespace Shopify {
...
   interface IInventoryItem {
     id: number;
     sku: string;
     tracked: boolean;
     created_at: string;
     updated_at: string;
     requires_shipping: boolean;
-    country_code_of_origin: string;
-    harmonized_system_code: string;
+    cost: string | null;
```
lpinca, line 1781:
```suggestion
    cost: string;
```

MBogdan18: This can be null if I remember correctly as you're not requested to set the cost per item and can be left blank. I'll look into it tomorrow and change the code if necessary.

MBogdan18: I have re-checked this and it can be indeed `null` if the cost per item hasn't been set.

lpinca: Ok, thank you.

lpinca (approved):

lpinca:
Thank you.

PR #443, opened 2020-12-05
Add GDPR webhook topics to types

joshgare:
Public apps are required to implement 3 GDPR web hook topics that are not included in `index.d.ts`.

https://shopify.dev/tutorials/add-gdpr-webhooks-to-your-app


hobbydevs:
Why was this not merged?

PR #447, opened 2020-12-19
This enables usage of pagination for orders and products

LarsBuur:
No original work here. It´s basically what @YourWishes suggests in #288.
Could and should be extended throughout the api for lists.

Usage example:
```
  const orderList: IOrder[] = [];
  const orderStatus = 'any';

  let params = { status: orderStatus, limit: 250 };

  do {
    const orders = await shopify.order.list(params);

    for (const order of orders) {
      orderList.push(order);
    }

    params = orders.nextPageParameters;
  } while (params !== undefined);

  console.log(`after while loop len:${orderList.length}`)
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -429,7 +429,7 @@ declare class Shopify {
     delete: (id: number) => Promise<any>;
     fulfillmentOrders: (id: number) => Promise<Shopify.IFulfillmentOrder[]>;
     get: (id: number, params?: any) => Promise<Shopify.IOrder>;
-    list: (params?: any) => Promise<Shopify.IOrder[]>;
+    list: (params?: any) => Promise<Shopify.IPaginatedResult<Shopify.IOrder>>;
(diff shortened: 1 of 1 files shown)


chaddjohnson:
I am not seeing a `nextPageParameters` property in response bodies when using a `limit` parameter.

It seems that Shopify has changed pagination to rely on response headers containing URL links to next pages; see [this article](https://shopify.dev/tutorials/make-paginated-requests-to-rest-admin-api).

dankeller101:
Just wanted to drop in to say this would be very helpful for us on our project that depends on this wrapper!  Let me know if there's any way I can help make this land sooner.

davidmm07:
@lpinca Check this PR

lpinca:
It should be done for all [supported endpoints](https://shopify.dev/tutorials/make-paginated-requests-to-rest-admin-api#supported-endpoints).

PR #455, opened 2021-02-01
Applied Discounts do not match Shopify documentation

anthlubic:
When adding an applied discount to a draft order or a draft order line item, you can only pass in a single discount.  Arrays don't seem to be accepted.

Reference: https://shopify.dev/docs/admin-api/rest/reference/orders/draftorder#create-2020-04


lpinca (approved):

lpinca:
Thank you.

PR #457, opened 2021-02-02
Iterator feature

chewax:
Added new iterator feature to the base mixin to be able to list and paginate at once. It will iterate though all elements that match `param` and handling pagination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -309,6 +309,7 @@ This feature is only available on version 2.24.0 and above.
   - `delete(id)`
   - `get(id[, params])`
   - `list([params])`
+  - `iterate([params])`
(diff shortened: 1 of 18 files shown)


lpinca:
Thanks but I'm not a fan of this. I don't like to have an `iterate` action when the ability to paginate results is [already available](https://github.com/MONEI/Shopify-api-node#pagination).

The current feature requires only a few more line of code and users that want to use `for await...of` can extend the classes like you did here.

chewax:
Ok no worries.
Have a great day.

lpinca:
I'm closing this. Thank you anyway.

PR #459, opened 2021-02-04
Add "cache" property to ScriptTag interfaces

ivy:
Hey folks - I updated the TypeScript definitions to include the ScriptTag `cache` property, as seen in the [docs](https://shopify.dev/docs/admin-api/rest/reference/online-store/scripttag) for API version 2021-01.


lpinca (approved):

lpinca:
Thank you.

PR #460, opened 2021-02-08
fix: ICollectionListing.collection_id definition

builtbylane:
ICollectionListing.collection_id is a number, not a string


lpinca (approved):

lpinca:
Thank you.

PR #461, opened 2021-02-11
[ts] Fix ITransaction["receipt"] typing

tkalliom:
The API specification states that the “value of `Transaction["receipt"]` depends on which gateway the shop is using”. It would appear that the current typings describe a receipt from Stripe.

I’m not totally certain what the best way to type this would be. At least with this type, it should possible to make typeguards for known gateways.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2782,14 +2782,6 @@ declare namespace Shopify {
   type TransactionSourceName = 'android' | 'iphone' | 'pos' | 'web';
   type TransactionStatus = 'error' | 'failure' | 'pending' | 'success';
 
-  interface ITRansactionReceipt {
-    testcase: boolean;
-    authorization: string;
-    balance_transaction?: {
-      [k: string]: any;
-    };
-  }
-
   interface ITransaction {
     amount: string;
     authorization: string;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2807,7 +2799,7 @@ declare namespace Shopify {
     payment_details: IPaymentDetails;
     parent_id: number;
     processed_at: string;
-    receipt: ITRansactionReceipt;
+    receipt: Record<string, unknown>;
```
lpinca, line 2802: I think this is not backwards compatible with TypeScript 2.

tkalliom: `Record` was introduced in TypeScript 2.1: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#partial-readonly-record-and-pick

lpinca: Yes but AFAIK `unknown` was introduced in TypeScript

tkalliom: Ah, you’re right. Changed to `any`

lpinca (approved):

lpinca:
Thank you.

PR #467, opened 2021-03-18
Fix require - use package.json instead of package

olliechick:


lpinca:
Thanks but this is not needed. See https://nodejs.org/api/modules.html#modules_file_modules.

olliechick:
I'm getting the error:

```
Error: ./node_modules/shopify-api-node/index.js
Module not found: Error: Can't resolve './package' in '/app/node_modules/shopify-api-node'
resolve './package' in '/app/node_modules/shopify-api-node'
  using description file: /app/node_modules/shopify-api-node/package.json (relative path: .)
    Field 'browser' doesn't contain a valid alias configuration
    using description file: /app/node_modules/shopify-api-node/package.json (relative path: ./package)
      no extension
        Field 'browser' doesn't contain a valid alias configuration
        /app/node_modules/shopify-api-node/package doesn't exist
      .ts
        Field 'browser' doesn't contain a valid alias configuration
        /app/node_modules/shopify-api-node/package.ts doesn't exist
      .tsx
        Field 'browser' doesn't contain a valid alias configuration
        /app/node_modules/shopify-api-node/package.tsx doesn't exist
      .mjs
        Field 'browser' doesn't contain a valid alias configuration
        /app/node_modules/shopify-api-node/package.mjs doesn't exist
      .js
        Field 'browser' doesn't contain a valid alias configuration
        /app/node_modules/shopify-api-node/package.js doesn't exist
      as directory
        /app/node_modules/shopify-api-node/package doesn't exist
[/app/node_modules/shopify-api-node/package]
[/app/node_modules/shopify-api-node/package.ts]
[/app/node_modules/shopify-api-node/package.tsx]
[/app/node_modules/shopify-api-node/package.mjs]
[/app/node_modules/shopify-api-node/package.js]
 @ ./node_modules/shopify-api-node/index.js 9:12-32
```

olliechick:
I fixed it by adding `config.resolve.extensions.push('.json')` to the `module.exports` of my webpack config (where `config` is the object returned)

PR #477, opened 2021-03-31
Implemented IPaginatedResult for supported endpoints

metasloth:
This is the continued/remaining effort to complete #447, and is also relevant to #288.

Resolves [this additional request](https://github.com/MONEI/Shopify-api-node/pull/447#issuecomment-809596052) by @lpinca:
> It should be done for all [supported endpoints](https://shopify.dev/tutorials/make-paginated-requests-to-rest-admin-api#supported-endpoints).

PR #479, opened 2021-04-01
Fixed nested array type error with new IPaginatedResults type

metasloth:
My recent PR #477 added redundant array type signature responses because I forgot how generics work 🤦 .

This makes typescript think that the promise results are an array of arrays, as seen below:
image

I removed all erroneous array signatures and actually tested the code in practice this time to make sure it works as expected!


lpinca (approved):

lpinca:
Thank you.

PR #481, opened 2021-04-08
[ts] Corrected ICarrierItem type definition.

YourWishes:
Carrier callback requests type from Shopify was expecting line item properties to be in format of;
`{ name: string, value: string }[]`
when the correct type is instead;
`{ [key: string]: string; }`


lpinca (approved):

lpinca:
Thank you.

PR #483, opened 2021-04-22
Fix type definitions

gabrielvincent:
This pull request fixes the following type definitions:

~`province: string`~  ->  `province: string | null`
~`province_code: string`~  ->  `province_code: string | null`
~`latitude: string`~  -> `latitude: number | null`
~`longitude string`~  -> `longitude: number | null` 

Example of actual data sent by the Shopify API:

types
_`province` can be `null`. `latitude`/`longitude` are `number`_

types2
_`latitude`/`longitude` can be `null`_

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1391,12 +1391,12 @@ declare namespace Shopify {
     first_name: string;
     id: number;
     last_name: string;
-    latitude: string;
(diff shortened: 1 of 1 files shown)


lpinca:
Thank you.

PR #486, opened 2021-05-05
chore (ts): 2021-04 order api additions

builtbylane:
see https://shopify.dev/changelog/duties-are-now-available-on-the-storefront-api
note: `current_total_duties_set` is slightly different than the rest of the additions:  there's no `current_total_duties` and the value `null` when duties are not applicable


lpinca (approved):

lpinca:
Thank you.

PR #497, opened 2021-07-21
bug: removed force_ssl according to new release 2021-07

samkit5495:
release information https://shopify.dev/api/release-notes/2021-07
Topic: Deprecated field on Shop resource 
force_ssl field was removed from Shop resource

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2711,7 +2711,6 @@ declare namespace Shopify {
     enabled_presentment_currencies: string[];
     email: string;
     finances: boolean;
-    force_ssl: boolean;
     google_apps_domain: string | null;
     google_apps_login_enabled: any | null;
     has_discounts: boolean;
--- a/index.d.ts
+++ b/index.d.ts
@@ -2711,6 +2711,7 @@ declare namespace Shopify {
     enabled_presentment_currencies: string[];
     email: string;
     finances: boolean;
+    force_ssl?: boolean;
     google_apps_domain: string | null;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -2711,7 +2711,6 @@ declare namespace Shopify {
     enabled_presentment_currencies: string[];
     email: string;
     finances: boolean;
-    force_ssl: boolean;
```
lpinca, line 2714: According to https://shopify.dev/api/admin/rest/reference/store-properties/shop the property was not removed but only deprecated. Its value is always `true`.

lpinca (changes requested):

lpinca (approved):
```js
import Shopify from 'shopify-api-node';

const shopify = new Shopify({
  shopName: 'quuz',
  apiVersion: '2021-07',
  apiKey: 'b68c3a4bf44cd240511e5f4114f0be4d',
  password: '2d170167fc62551920ca60fe40a5e548'
});

const info = await shopify.shop.get();
console.log(info.force_ssl);
```

index.d.ts
samkit5495: ok so can we make it not mandatory

lpinca: Yes, please do that.

lpinca (approved):

index.d.ts
samkit5495: done

lpinca:
Thank you.

Issue #500, opened 2021-08-09
BigInt ID handling 


airhorns:
Shopify's REST APIs produce ID numbers in the JSON responses that require more than 53 bits of precision to represent, which when parsed with the plain old `JSON.parse` built in lose precision, resulting in issues like these: https://community.shopify.com/c/Shopify-APIs-SDKs/Int-too-big-to-fit-in-my-API/td-p/1070054

As best I can tell, this library uses `JSON.stringify` and `JSON.parse` (via got) to send and receive JSON, which means that the id numbers sent and returned from various calls will be incorrect! I hope I am missing something, but otherwise, I think it would be good for this library to do something to help users properly parse (and send) bigints to/from Shopify.

In the past, solutions like [`json-bigint`](https://github.com/sidorares/json-bigint) used userland implementations of arbitrary precision numbers like `big.js`, but now, newer ES versions have the `bigint` type built in. `json-bigint` allows that type to be used natively for parsing and stringifying with the [`useNativeBigint`](https://github.com/sidorares/json-bigint#optionsusenativebigint-boolean-default-false) option, so would it make sense to switch to it for all that and ask users to use native bigints to represent big IDs?

lpinca:
To be honest I think this is a Shopify issue. Other APIs use strings for 64 bit ids because most JSON parsers use the same precision as the IEEE 754 double.

Anyway this should eventually be opened in `got` issue tracker as JSON parsing and serialization is done by `got`. We call [`JSON.stringify()`](https://github.com/MONEI/Shopify-api-node/blob/3.6.14/index.js#L228) explicitly but only for GraphQL.

airhorns:
> To be honest I think this is a Shopify issue.

Yeah, I don't disagree, but I don't think they are gonna change it any time soon. I used to work there and when they switched to 64 bit ints internally in like 2014 they specifically decided not to do strings as output because it would cause such a huge amount of downstream pain for everyone, and a lot of not-JS languages (like Ruby) parse big numbers in JSON just fine. I agree it is super annoying though.

> Anyway this should eventually be opened in the got issue tracker as JSON parsing and serialization is done by got. 

Yeah, I think if you peel back the onion enough, it really should be the default behaviour for `JSON.parse` and `JSON.stringify` if you ask me, so it works for `got` and `axios` and `node-fetch` and and and and. But, TC-39 hasn't yet decided on it I don't think. In the meantime, are you open to fixes that make it better for this API specifically which we know emits big ints?

[...]

PR #502, opened 2021-08-10
Add ability to use a custom JSON parser and serializer

lpinca:
Fixes: https://github.com/MONEI/Shopify-api-node/issues/500

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -6,7 +6,7 @@ extends:
   - eslint:recommended
   - plugin:prettier/recommended
 parserOptions:
-  ecmaVersion: 9
+  ecmaVersion: latest
 rules:
   no-var: error
   prefer-const: error
--- a/README.md
+++ b/README.md
@@ -54,8 +54,16 @@ Creates a new `Shopify` instance.
   specifies a limit of 2 requests per second with a burst of 35 requests. When
   set to `true` requests are limited as specified in the above example. Defaults
   to `false`.
+- `parseJson` - Optional - The function used to parse JSON. The function is
+  passed a single argument. This option allows the use of a custom JSON parser
+  that might be needed to properly handle long integer IDs. Defaults to
+  `JSON.parse()`.
 - `presentmentPrices` - Optional - Whether to include the header to pull
   presentment prices for products. Defaults to `false`.
+- `stringifyJson` - Optional - The function used to serialize to JSON. The
+  function is passed a single argument. This option allows the use of a custom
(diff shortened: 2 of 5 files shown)


lpinca:
cc: @airhorns

airhorns:
Nice! Do you think it makes sense to set `json-bigint` as the defaults since we know that Shopify is gonna return big integers? Make it work by default for users so they don't have to care? Without the defaults, I feel like there will be a stream of sad people who get hit with this very confusing error in production and have to go spelunking to find the issue and the remedy and then set the default to `json-bigint` which seems like a bit of wasted effort. 

It would change the minimum node version from `10.0` to `10.4` which doesn't seem too bad to me, but I think strictly speaking it'd also be a breaking change since downstream callers might not expect bigints to come back.

lpinca:
> but I think strictly speaking it'd also be a breaking change since downstream callers might not expect bigints to come back.

Yes, this is why I prefer to use `JSON.parse()` and `JSON.stringify()` as default and let users use a custom parser/serializer if they need to. It would also be easier for users who wants to experiment with https://github.com/tc39/proposal-json-parse-with-source. without relying on `json-bigint` or another dependency.

lpinca:
I've tried to make `json-bigint` `parse()` and the `stringify()` the defaults but I hit this bug https://github.com/sidorares/json-bigint/issues/49 for https://github.com/MONEI/Shopify-api-node/blob/3.6.14/test/fixtures/draft-order/res/list.json#L59.

`json-bigint` is currently a no go. Do you know any other safe alternative? We can also try to fix that issue but I'm not sure if `json-bigint` is actively maintained.

airhorns:
That is a big issue with `json-bigint`, yeesh. No go indeed. I don't know of any alternatives which again seems nuts to me that tonnes of people aren't hitting this problem ... but alas. I will see if I can upstream a fix to `json-bigint` or dig something else up!

lpinca:
There is also another issue with `json-bigint`. If a number has more than 15 digits (including the dot for floats) it will parse it using bignumber.js / BigInt even if it is not needed. For example, there is no need to use bignumber.js for -75.69314750000001 and I expect it to be parsed as a plain JS number.

lpinca:
I think I'll merge this as is and let users deal with these issues themselves.

weeksie:
This is horrifying. The answer to client developers getting silently wrong values is to shrug and say I dunno, we couldn't figure it out, you do it. 

I was just going to move over from doing raw fetch/rest calls but I guess I won't now? The Shopify ecosystem has been a non stop parade of surprises, that's for sure

PR #525, opened 2022-02-03
[ts] Fix IOrderShippingLine type definition

zronn:
This pull request fixes missing `id` attribut on `IOrderShippingLine`.

Here is an example from Shopify response :

screeshot


lpinca (approved):

lpinca:
Thank you.

PR #541, opened 2022-06-21
Add an option for automatic retries when hitting rate limits

### LINKED ISSUE #251: Adding retry option
First, thanks for your work on this module!

I found a related issue which was immediately closed for some reason by its author: #238 

More and more frequently the Shopify API is returning ETIMEDOUT

I noticed a `timeout` option was made available to be passed to the got request in pr #98.  Are there plans to add an option leverage got's `retry` option in the same manner?  Would you accept a pull request in that regard?

airhorns:
This adds opt-in retry logic to calls made to REST or GraphQL endpoints on Shopify. Many users of the Shopify API run into the restrictive rate limits, and there's one simple way you can work around them, which is to retry the request after the time has elapsed. This isn't the only strategy, but it's a really common one because it doesn't require any cross-process coordination or state. Instead of trying to guess before hand if a request is going to work or not (which is hard to do if you have many different processes on many different servers all making requests at the same time), this will just always try the request, and keep retrying for a while until it succeeds. 

This approach uses Got's built in `retry` configuration option, so `shopify-api-node` doesn't have to implement any of the custom backoff logic or actual retry code, and instead just has to compute the time to wait before retrying which is kind of nice.

I think Shopify's restrictive rate limits burn just about every user of the Shopify APIs really soon after they get started using it -- the limits are pretty much the lowest in the business, and are a big part of working with the Shopify API. It's rare in my experience that you *don't* need to care about rate limiting, and so I think it'd be really great to have a safe-by-default rate limiting work around built into this library since many many users need something. I agree that a rate limit error is technically speaking a user error, but, why force every user to work around this error with bespoke code copied from a github issue? It's so common that I think it's worth building in. This approach is simple and robust (and used in the hosted Shopify sync product I'm working on), and I'd love for everyone else using this lib to not have to reinvent this every time. I took the time to try to make this good quality code that's tested to ease the maintenance burden as well.

I didn't turn on this behaviour by default so as to not make a breaking change, but for what it's worth, I think I will always turn it on when using this library in the future.  We could consider turning it on by default in the next major version of `shopify-api-node`, but that's up to the maintainers of course! 

Closes #251, references #324 and #199 and #377

### DIFF
--- a/index.js
+++ b/index.js
@@ -159,63 +159,60 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify retry-after header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay: calculateRESTDelay
+            calculateDelay
           }
         : 0,
-    method
+    method,
+    hooks: {
+      afterResponse: [
+        (res) => {
+          this.updateLimits(res.headers['x-shopify-shop-api-call-limit']);
+          return res;
+        }
+      ]
+    }
   };
 
   if (data) {
     options.json = key ? { [key]: data } : data;
   }
 
-  return got(uri, options).then(
-    (res) => {
-      const body = res.body;
+  return got(uri, options).then((res) => {
+    const body = res.body;
 
-      this.updateLimits(res.headers['x-shopify-shop-api-call-limit']);
+    if (res.statusCode === 202 && res.headers['location']) {
+      const retryAfter = res.headers['retry-after'] * 1000 || 0;
+      const { pathname, search } = url.parse(res.headers['location']);
 
-      if (res.statusCode === 202 && res.headers['location']) {
-        const retryAfter = res.headers['retry-after'] * 1000 || 0;
-        const { pathname, search } = url.parse(res.headers['location']);
+      return delay(retryAfter).then(() => {
+        const uri = { pathname, ...this.baseUrl };
 
-        return delay(retryAfter).then(() => {
-          const uri = { pathname, ...this.baseUrl };
+        if (search) uri.search = search;
 
-          if (search) uri.search = search;
+        return this.request(uri, 'GET', key);
+      });
+    }
 
-          return this.request(uri, 'GET', key);
+    const data = key ? body[key] : body || {};
+
+    if (res.headers.link) {
+      const link = parseLinkHeader(res.headers.link);
+
+      if (link.next) {
+        Object.defineProperties(data, {
+          nextPageParameters: { value: link.next.query }
         });
       }
 
-      const data = key ? body[key] : body || {};
-
-      if (res.headers.link) {
-        const link = parseLinkHeader(res.headers.link);
-
-        if (link.next) {
-          Object.defineProperties(data, {
-            nextPageParameters: { value: link.next.query }
-          });
-        }
-
-        if (link.previous) {
-          Object.defineProperties(data, {
-            previousPageParameters: { value: link.previous.query }
-          });
-        }
+      if (link.previous) {
+        Object.defineProperties(data, {
+          previousPageParameters: { value: link.previous.query }
+        });
       }
-
-      return data;
-    },
-    (err) => {
-      this.updateLimits(
-        err.response && err.response.headers['x-shopify-shop-api-call-limit']
-      );
-
-      return Promise.reject(err);
     }
-  );
+
+    return data;
+  });
 };
 
 /**
@@ -262,7 +259,6 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
   const uri = { pathname, ...this.baseUrl };
   const json = variables !== undefined && variables !== null;
-  let attemptNumber = 0;
   const options = {
     headers: {
       ...this.baseHeaders,
@@ -279,65 +275,45 @@ Shopify.prototype.graphql = function graphql(data, variables) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify retry-after header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay: (retryObject) => {
-              attemptNumber += 1;
-              return (
-                graphQLRetryMS(
-                  attemptNumber,
-                  this.options.maxRetries,
-                  retryObject.error
-                ) || retryObject.computedValue
-              );
-            }
+            calculateDelay
           }
         : 0,
     hooks: {
-      afterResponse:
-        this.options.maxRetries > 0
-          ? [
-              (response) => {
-                if (response.body && response.body.errors) {
-                  const retryDelay = graphQLRetryMS(
-                    attemptNumber,
-                    this.options.maxRetries,
-                    { response }
-                  );
+      afterResponse: [
+        (response) => {
+          if (response.body) {
+            if (response.body.extensions && response.body.extensions.cost) {
+              this.updateGraphqlLimits(response.body.extensions.cost);
+            }
 
-                  if (retryDelay) {
-                    // Make got consider this response errored and retry
-                    throw new Error('GraphQL cost limit exceeded, retrying');
-                  }
-                }
+            if (response.body.errors) {
+              // Make got consider this response errored and retry if needed
+              throw new Error(response.body.errors[0].message);
+            }
+          }
 
-                return response;
-              }
-            ]
-          : []
+          return response;
+        }
+      ],
+      beforeError: [errorForGraphQLError]
     }
   };
 
-  return got(uri, options).then((res) => {
-    if (res.body.extensions && res.body.extensions.cost) {
-      this.updateGraphqlLimits(res.body.extensions.cost);
-    }
-
-    if (res.body.errors) {
-      const first = res.body.errors[0];
-      const err = new Error(first.message);
-
-      err.locations = first.locations;
-      err.path = first.path;
-      err.extensions = first.extensions;
-      err.response = res;
-
-      throw err;
-    }
-
-    return res.body.data || {};
-  });
+  return got(uri, options).then(returnResponseData);
 };
 
 resources.registerAll(Shopify);
+
+/**
+ * Return the data of a GraphQL response object
+ *
+ * @param {Response} res Got response object
+ * @return {Object} The data
+ * @private
+ */
+function returnResponseData(res) {
+  return res.body.data;
+}
 
 /**
  * Returns a promise that resolves after a given amount of time.
@@ -438,29 +414,31 @@ function maybeRetryMS(error) {
 }
 
 /**
- * Return a duration after which to retry a graphql call, or null if it
- * shouldn't be retried
- *
- * @param {Number} attemptNumber Shopify graphql call attempt no
- * @param {Number} maxRetries Max retries in default option
- * @param {Object} error Error object from got call
- * @return {Boolean | undefined}
+ * Decorate a ResponseError object thrown by got with details from
+ * GraphQL errors in the response body
+ * @param {ResponseError} error
+ * @returns ResponseError
  * @private
  */
-function graphQLRetryMS(attemptNumber, maxRetries, error) {
-  if (attemptNumber <= maxRetries) {
-    return maybeRetryMS(error);
+function errorForGraphQLError(error) {
+  if (error.response && error.response.body.errors) {
+    const first = error.response.body.errors[0];
+
+    error.locations = first.locations;
+    error.path = first.path;
+    error.extensions = first.extensions;
   }
+  return error;
 }
 
 /**
- * Got `calculateDelay` hook function passed when calling the REST API
+ * Got `calculateDelay` hook function passed to decide how long to wait before retrying
  *
  * @param {Object} retryObject got's input for the retry logic
  * @return {Number}
  * @private
  */
-function calculateRESTDelay(retryObject) {
+function calculateDelay(retryObject) {
   return maybeRetryMS(retryObject.error) || retryObject.computedValue;
 }
 


package.json
```diff
@@ -35,7 +35,7 @@
     "prettier": "^2.0.2"
   },
   "scripts": {
-    "test": "c8 --reporter=lcov --reporter=text mocha",
+    "test": "c8 --reporter=lcov --reporter=text mocha --timeout 15000",
```
airhorns, line 38: I can switch the tests to use a fake timer implementation if this is distasteful -- wasn't sure what would be preferred

lpinca:
This should be mutually exclusive with the `autoLimit` option as it does not address https://github.com/MONEI/Shopify-api-node/issues/251#issuecomment-487965284. Also retrying on 5xx and network errors is questionable.

I would prefer something simpler where only 429 errors are retried if the `autoLimit` option is not set similarly to how it is done for the 202 responses.

index.js
```diff
@@ -313,4 +380,116 @@ function reducer(acc, cur) {
...
+  const response = responseFromError(error);
+  if (!response) {
+    return null;
+  }
+
+  if (response.headers && response.headers['retry-after']) {
+    const value = parseFloat(response.headers['retry-after']);
+    if (isNaN(value)) {
+      const when = new Date(value).valueOf();
+      return when - Date.now().valueOf();
```
lpinca, line 437: This returns `NaN`.

index.js
```diff
@@ -313,4 +380,116 @@ function reducer(acc, cur) {
...
+    } else if (isFinite(value)) {
+      return value;
+    } else {
+      return null;
+    }
+  }
+
+  if (
+    response.statusCode == 429 ||
```
lpinca, line 446:
```suggestion
    response.statusCode === 429 ||
```

airhorns:
> This should be mutually exclusive with the autoLimit option as it does not address https://github.com/MONEI/Shopify-api-node/issues/251#issuecomment-487965284. 

Ah right! I will update to throw in the constructor if both options are engaged.

> Also retrying on 5xx and network errors is questionable.

Can you explain your hesitance a bit more here? A 500 error from Shopify is Shopify telling the client that Shopify is at fault and the client didn't do anything wrong. I think that means means that there's really nothing to do but retry. In my experience Shopify throws transient errors like this all the time and it seems quite convenient to not have to wrap `shopify-api-node` in yet some other retry library to catch one class of errors. Is there something you think users should do with those errors other than retry? Also there's prior art for this, got's default is 2 retries and automatic retrying of 5xx series responses as well as network errors: https://github.com/sindresorhus/got/blob/main/documentation/7-retry.md#retry . Also fwiw, here's what we see in production across all our shopify apps for shopify response codes -- enough 503s in there that this makes sense to me:
CleanShot 2022-06-22 at 14 05 58@2x

index.js
lpinca: What about other 4xx that Got retires?

index.js
```diff
@@ -313,4 +380,116 @@ function reducer(acc, cur) {
...
+    return 2 * 1000;
+  }
+
+  // detect graphql request throttling
+  if (response.body && typeof response.body === 'object') {
+    const body = response.body;
+
+    if (
+      body.errors &&
+      Array.isArray(body.errors) &&
```
lpinca, line 459: Isn't `body.errors` always an array?

index.js
```diff
@@ -313,4 +380,116 @@ function reducer(acc, cur) {
...
+      typeof body.errors[0] === 'object' &&
+      body.errors[0].extensions?.code == 'THROTTLED'
```
lpinca, line 461: This does not work on Node.js 10

test/shopify.test.js
```diff
@@ -529,9 +529,158 @@ describe('Shopify', () => {
...
+
+    const shopifyRetryableInstance = new Shopify({
+      accessToken,
+      parseJson,
+      shopName,
+      stringifyJson,
+      maxRetries: 3
+    });
+
+    it('should retry 429 errors from Shopify after the retry-after header has elapsed', async () => {
```
lpinca, line 552: Remove "should" and use third person present tense: `it('retries 429...'` and try to make it fit in an 80 chars line. It would also be nice to be consistent with other tests and use `async...await` only if it greatly simplifies the tests.

test/shopify.test.js
```diff
@@ -529,9 +529,158 @@ describe('Shopify', () => {
...
+      const shopify = shopifyRetryableInstance;
+
+      nock(`https://${shopName}.myshopify.com`)
```
lpinca, line 555: Use `scope`.

test/shopify.test.js
```diff
@@ -529,9 +529,158 @@ describe('Shopify', () => {
...
+        .get('/admin/shop.json')
+        .reply(429, () => 'too many requests', { 'Retry-After': '1' })
+        .get('/admin/shop.json')
+        .reply(429, () => 'too many requests', { 'Retry-After': '2' });
+
+      addWorkingRESTRequestMock();
```
lpinca, line 561: Pass `scope` to this function and use it there.

test/shopify.test.js
```diff
@@ -529,9 +529,158 @@ describe('Shopify', () => {
...
+
+    const addWorkingRESTRequestMock = () => {
+      nock(`https://${shopName}.myshopify.com`)
+        .get('/admin/shop.json')
+        .reply(200, () => ({
+          shop: {
+            name: 'My Cool Test Shop',
+            id: '1'
+          }
+        }));
```
lpinca, line 541:
```suggestion
        .reply(200, {
          shop: {
            name: 'My Cool Test Shop',
            id: '1'
          }
        });
```

For consistency with other tests, please do not use functions for the second argument of `reply()` .

index.js
```diff
@@ -23,8 +34,8 @@ const resources = require('./resources');
  * @param {Boolean|Object} [options.autoLimit] Limits the request rate
  * @param {Number} [options.timeout] The request timeout
  * @param {Function} [options.parseJson] The function used to parse JSON
- * @param {Function} [options.stringifyJson] The function used to serialize to
- *     JSON
+ * @param {Function} [options.stringifyJson] The function used to serialize to JSON
+ * @param {Number} [options.maxRetries] Maximum number of automatic request retries, defaults to 0 (no retries)
```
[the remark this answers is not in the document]
lpinca: Yes, please.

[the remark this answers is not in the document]
lpinca: Anyway this is weird. I've just run

```
prettier --write --ignore-path .gitignore "**/*.{json,md,ts,yaml,yml}"
```

and nothing has changed.

index.js
```diff
@@ -235,6 +258,7 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
   const uri = { pathname, ...this.baseUrl };
   const json = variables !== undefined && variables !== null;
+  let attemptNumber = 0;
```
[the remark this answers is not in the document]
airhorns: Updated to use the same counter for both classes of error

index.js
airhorns: It is but since it is coming from the outside world I think it is prudent to not assume the exact structure, especially when handling errors?

index.js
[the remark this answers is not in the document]
airhorns: I think I was accidentally using a version of prettier from my environment -- my bad. Fixed.

index.js
```diff
@@ -243,12 +276,52 @@ Shopify.prototype.graphql = function graphql(data, variables) {
...
-    body: json ? this.options.stringifyJson({ query: data, variables }) : data
+    body: json ? this.options.stringifyJson({ query: data, variables }) : data,
+    retry:
+      this.options.maxRetries > 0
+        ? {
+            limit: this.options.maxRetries,
+            // Don't clamp Shopify retry-after header values too low.
+            maxRetryAfter: Infinity,
+            calculateDelay: (retryObject) => {
+              attemptNumber += 1;
```
lpinca, line 288: I think this does not solves the issue. For example assume that a request is answered with 200 but has errors (`request.body.errors`). Right now it does not even increment `attemptNumber` but more importantly even if it does, that counter is not used when the request is answered, for example, with 500. So if the limit is 3 and the request is answered with 200+error, 200+error, 500, 418, the limit might not be respected.

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+  if (
+    typeof error === 'object' &&
+    'code' in error &&
+    RetryableErrorCodes.has(error.code)
+  ) {
+    return 1000;
+  }
+
+  const response = error.response;
+  if (!response || !response.headers) {
```
lpinca, line 408:
```suggestion
  if (!response) {
```

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+ * @return {Boolean | null}
+ * @private
+ **/
+function maybeRetryMS(error) {
+  // for simplicity, retry network connectivity issues after a hardcoded 1s
+  if (
+    typeof error === 'object' &&
+    'code' in error &&
+    RetryableErrorCodes.has(error.code)
+  ) {
```
lpinca, line 403:
```suggestion
  if (RetryableErrorCodes.has(error.code)) {
```

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+  const response = error.response;
+  if (!response || !response.headers) {
+    return null;
+  }
+
+  if (response.headers && response.headers['retry-after']) {
+    const value = parseFloat(response.headers['retry-after']);
+    if (isNaN(value)) {
+      const when = new Date(value).valueOf();
+      return when - Date.now();
```
lpinca, line 416: This still returns `NaN` because `Number.isNaN(new Date(NaN).valueOf()) === true.`

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+    } else if (isFinite(value)) {
+      return value;
```
lpinca, line 418:
```suggestion
      return value * 1000;
```

Wouldn't it be simpler to remove `parseFloat` and all the check and just return `response.headers['retry-after'] * 1000 || null`?

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+      return null;
+    }
+  }
+
+  if (
+    response.statusCode === 429 ||
+    response.statusCode === 418 ||
+    response.statusCode === 413 ||
+    (response.statusCode >= 500 && response.statusCode < 600)
+  ) {
```
lpinca, line 429:
```suggestion
  if (response.statusCode < 200 || response.statusCode > 300) {
```

index.js
lpinca: Something similar is done in `Shopify.prototype.graphql()` and we are skipping this and the following (`typeof body.errors[0] === 'object'`) check.

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+    const body = response.body;
+
+    if (
+      body.errors &&
+      Array.isArray(body.errors) &&
+      typeof body.errors[0] === 'object' &&
+      body.errors[0].extensions &&
+      body.errors[0].extensions.code == 'THROTTLED'
+    ) {
+      const costData = body.extensions?.cost;
```
lpinca, line 447:
```suggestion
      const costData = body.extensions && body.extensions.cost;
```

index.js
```diff
@@ -313,4 +386,95 @@ function reducer(acc, cur) {
...
+}
+
+/**
+ * Return a duration after which to retry a graphql call, or null if it
+ * shouldn't be retried
+ *
+ * @param {Number} attemptNumber Shopify graphql call attempt no
+ * @param {Number} maxRetries Max retries in default option
+ * @param {Object} error Error object from got call
+ * @return {Boolean | null}
```
lpinca, line 471:
```suggestion
 * @return {Boolean | undefined}
```

test/shopify.test.js
```diff
@@ -123,7 +133,7 @@ describe('Shopify', () => {
     const url = { pathname: '/test', ...shopify.baseUrl };
     const scope = common.scope;
 
-    afterEach(() => expect(nock.isDone()).to.be.true);
+    afterEach(() => expect(nock.pendingMocks()).to.deep.equal([]));
```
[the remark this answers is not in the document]
lpinca: It might indeed be useful but the same pattern is used everywhere so for consistency I would keep it as is for now and possibly change it later everywhere in another PR.

test/shopify.test.js
```diff
@@ -835,5 +978,127 @@ describe('Shopify', () => {
...
+
+      addWorkingGraphQLRequestMock(scope);
+      return shopify.graphql('query { shop { id name } }').then((result) => {
+        expect(result.shop.name).equal('My Cool Test Shop');
+      });
+    });
+
+    it('retries network system level errors immediately', () => {
+      scope.post('/admin/api/graphql.json').replyWithError({
+        message: 'the network borken',
```
lpinca, line 1035:
```suggestion
        message: 'the network is broken',
```

index.js
```diff
@@ -44,9 +57,16 @@ function Shopify(options) {
     parseJson: JSON.parse,
     stringifyJson: JSON.stringify,
     timeout: 60000,
+    maxRetries: 0,
     ...options
   };
 
+  if (this.options.autoLimit && this.options.maxRetries > 0) {
```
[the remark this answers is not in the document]
airhorns: I've actually found the unspecificness of this constructor error message really unhelpful -- when a developer sees the other message, they are given no guidance on how to fix the issue, or which of the 4 issues actually occurred. They need to somehow reconstruct the state passed to the constructor to understand how to fix it, when the constructor already actually knows exactly what's wrong. I can do it in a different PR if you like but I think specific error messages would be a lot more helpful!

index.js
airhorns: Ah I messed this up. I will fix it -- it's supposed to date parse the It was actually supposed to be an attempt to parse timestamps (not durations) returned in the `Retry-After` header, which was a thing we observed Shopify do at one point. They don't document that they do that anywhere so I can remove this if you wish, but we definitely saw them do it at some point.

index.js
airhorns: I don't think we want to retry 404 errors or 401 errors which is why Got is selective. This matches their defaults

index.js
lpinca: > When you call `retryWithUpdatedOptions` in that after response hook, it still comes through this code path, as it is a retry, which I think is smart of got to do that way.

I did not know this. I think it makes sense now. Let me think a bit more about correctness.

index.js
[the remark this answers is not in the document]
lpinca: It makes sense but it is currently inconsistent with the other options. There are a total of 6 options one of which is mandatory so it shouldn't be that hard to read the documentation and fix the issue. I would prefer a separate PR.

index.js
lpinca: Ok, I see that problem here is that `maybeRetryMS()` can actually be called for status codes that got does not retries due its usage in GraphQL requests, but in this case we should be more stringent also on 5xx. See https://github.com/sindresorhus/got/tree/v11.8.5#retry.

index.js
```diff
@@ -313,4 +386,91 @@ function reducer(acc, cur) {
...
+    return null;
+  }
+
+  if (response.headers && response.headers['retry-after']) {
+    const value = parseFloat(response.headers['retry-after']);
+    if (isNaN(value)) {
+      const when = new Date(response.headers['retry-after']);
+      if (!isNaN(when)) {
+        return when.valueOf() - Date.now();
+      }
```
lpinca, line 414: I think an else branch is missing here.

index.js
```diff
@@ -243,12 +275,52 @@ Shopify.prototype.graphql = function graphql(data, variables) {
...
+            }
+          }
+        : 0
   };
 
-  return got(uri, options).then((res) => {
+  const instance = got.extend({
+    hooks: {
+      afterResponse: [
+        (response, retryWithMergedOptions) => {
```
lpinca, line 303: I think this hook should not be added when the `autoLimit` option is set. Also, instead of using `got.extend()` isn't it possible to add the `hooks` object in the `options` object?

airhorns: I tried as well to get rid of the `got.extend`, but you need a reference to the instance to grab the options to pass to the `retryWithMergedOptions` function. It pukes if you don't send in the same instance's options or no options at all. This `instance =` pattern is copied from the example here: https://github.com/sindresorhus/got/blob/main/documentation/9-hooks.md#afterresponse

lpinca: Hmm even if you pass an empty options object?

index.js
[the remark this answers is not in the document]
lpinca: Only one test is failing with this:

```diff
$ git diff
diff --git a/index.js b/index.js
index 82a4a0f..5bd8873 100644
--- a/index.js
+++ b/index.js
@@ -297,33 +297,29 @@ Shopify.prototype.graphql = function graphql(data, variables) {
         : 0
   };
 
-  const instance = got.extend({
-    hooks: {
-      afterResponse:
-        this.options.maxRetries > 0
-          ? [
-              (response, retryWithMergedOptions) => {
-                if (response.body && response.body.errors) {
-                  const retryDelay = graphQLRetryMS(
-                    attemptNumber,
-                    this.options.maxRetries,
-                    { response }
-                  );
-
-                  if (retryDelay) {
-                    return retryWithMergedOptions(instance.default.options);
-                  }
-                }
-
-                return response;
-              }
-            ]
-          : []
-    },
-    mutableDefaults: true
-  });
+  if (this.options.maxRetries > 0) {
+    options.hooks = {
+      afterResponse: [
+        (response, retryWithMergedOptions) => {
+          if (response.body && response.body.errors) {
+            const retryDelay = graphQLRetryMS(
+              attemptNumber,
+              this.options.maxRetries,
+              { response }
+            );
+
+            if (retryDelay) {
+              return retryWithMergedOptions(options);
+            }
+          }
+
+          return response;
+        }
+      ]
+    };
+  }
 
-  return instance(uri, options).then((res) => {
+  return got(uri, options).then((res) => {
     if (res.body.extensions && res.body.extensions.cost) {
       this.updateGraphqlLimits(res.body.extensions.cost);
     }
```

index.js
lpinca: Only one test fails with this:

```diff
$ git diff
diff --git a/index.js b/index.js
index 82a4a0f..5bd8873 100644
--- a/index.js
+++ b/index.js
@@ -297,33 +297,29 @@ Shopify.prototype.graphql = function graphql(data, variables) {
         : 0
   };
 
-  const instance = got.extend({
-    hooks: {
-      afterResponse:
-        this.options.maxRetries > 0
-          ? [
-              (response, retryWithMergedOptions) => {
-                if (response.body && response.body.errors) {
-                  const retryDelay = graphQLRetryMS(
-                    attemptNumber,
-                    this.options.maxRetries,
-                    { response }
-                  );
-
-                  if (retryDelay) {
-                    return retryWithMergedOptions(instance.default.options);
-                  }
-                }
-
-                return response;
-              }
-            ]
-          : []
-    },
-    mutableDefaults: true
-  });
+  if (this.options.maxRetries > 0) {
+    options.hooks = {
+      afterResponse: [
+        (response, retryWithMergedOptions) => {
+          if (response.body && response.body.errors) {
+            const retryDelay = graphQLRetryMS(
+              attemptNumber,
+              this.options.maxRetries,
+              { response }
+            );
+
+            if (retryDelay) {
+              return retryWithMergedOptions(options);
+            }
+          }
+
+          return response;
+        }
+      ]
+    };
+  }
 
-  return instance(uri, options).then((res) => {
+  return got(uri, options).then((res) => {
     if (res.body.extensions && res.body.extensions.cost) {
       this.updateGraphqlLimits(res.body.extensions.cost);
     }
```

index.js
```diff
@@ -243,12 +275,55 @@ Shopify.prototype.graphql = function graphql(data, variables) {
...
+                  retryObject.error
+                ) || retryObject.computedValue
+              );
+            }
+          }
+        : 0
   };
 
-  return got(uri, options).then((res) => {
+  const instance = got.extend({
```
lpinca, line 300: See https://github.com/MONEI/Shopify-api-node/pull/541#discussion_r905397960

index.js
```diff
@@ -134,7 +153,19 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
...
+    retry:
+      this.options.maxRetries > 0
+        ? {
+            limit: this.options.maxRetries,
+            // Don't clamp Shopify retry-after header values too low.
+            maxRetryAfter: Infinity,
+            calculateDelay: (retryObject) => {
+              return (
+                maybeRetryMS(retryObject.error) || retryObject.computedValue
+              );
```
lpinca, line 165: Can you please move this at the root level? The function does not depend on anything defined on this scope, so there is no need to redefine it every time.

index.js
airhorns: Nice catch -- definitely not on purpose. Instead of trying to figure out what strange options we need to pass to that retry function, I updated this to throw, which I actually think matches what is happening better. We've got a response but we really did encounter an error condition (if this mode is configured), so throwing seems like a nice and explicit way of triggering a retry. I tried to get that retry function working but it requires the ugly `instance =` stuff and an options merge that seems more complicated than it is worth.

index.js
```diff
@@ -243,9 +271,49 @@ Shopify.prototype.graphql = function graphql(data, variables) {
...
+            }
+          }
+        : 0,
+    hooks: {
+      afterResponse:
+        this.options.maxRetries > 0
+          ? [
+              (response) => {
+                if (response.body && response.body.errors) {
+                  const retryDelay = graphQLRetryMS(
```
lpinca, line 300: I think `graphQLRetryMS()` is not needed. If I'm not wrong we can use `maybeRetryMS()` directly instead, and remove `attemptNumber`. `got` will handle the number of retries internally. `maybeRetryMS()` can return `null` if the HTTP response has an invalid `Retry-After` header but I hope such case (`response.body.error` + invalid `Retry-After`) is not possible.

index.js
```diff
@@ -243,9 +271,49 @@ Shopify.prototype.graphql = function graphql(data, variables) {
...
+                    attemptNumber,
+                    this.options.maxRetries,
+                    { response }
+                  );
+
+                  if (retryDelay) {
+                    // Make got consider this response errored and retry
+                    throw new Error('GraphQL cost limit exceeded, retrying');
```
lpinca, line 308: I would create an error like it is done below starting on line 325, so that error is used if the maximum number of attempts is reached.

index.js
lpinca: I'm also not very happy to call `maybeRetryMS()` twice, here and in `calculateDelay()` to compute the same value twice but I have no better idea at the moment.

lpinca: I mean something like this

```diff
diff --git a/index.js b/index.js
index 10f8500..c2a8610 100644
--- a/index.js
+++ b/index.js
@@ -159,9 +159,17 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify retry-after header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay: calculateRESTDelay
+            calculateDelay
           }
         : 0,
+    hooks: {
+      afterResponse: [
+        (res) => {
+          this.updateLimits(res.headers['x-shopify-shop-api-call-limit']);
+          return res;
+        }
+      ]
+    },
     method
   };
 
@@ -173,8 +181,6 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
     (res) => {
       const body = res.body;
 
-      this.updateLimits(res.headers['x-shopify-shop-api-call-limit']);
-
       if (res.statusCode === 202 && res.headers['location']) {
         const retryAfter = res.headers['retry-after'] * 1000 || 0;
         const { pathname, search } = url.parse(res.headers['location']);
@@ -262,7 +268,6 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
   const uri = { pathname, ...this.baseUrl };
   const json = variables !== undefined && variables !== null;
-  let attemptNumber = 0;
   const options = {
     headers: {
       ...this.baseHeaders,
@@ -279,48 +284,44 @@ Shopify.prototype.graphql = function graphql(data, variables) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify retry-after header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay: (retryObject) => {
-              attemptNumber += 1;
-              return (
-                graphQLRetryMS(
-                  attemptNumber,
-                  this.options.maxRetries,
-                  retryObject.error
-                ) || retryObject.computedValue
-              );
-            }
+            calculateDelay
           }
         : 0,
     hooks: {
-      afterResponse:
-        this.options.maxRetries > 0
-          ? [
-              (response) => {
-                if (response.body && response.body.errors) {
-                  const retryDelay = graphQLRetryMS(
-                    attemptNumber,
-                    this.options.maxRetries,
-                    { response }
-                  );
-
-                  if (retryDelay) {
-                    // Make got consider this response errored and retry
-                    throw new Error('GraphQL cost limit exceeded, retrying');
-                  }
-                }
-
-                return response;
-              }
-            ]
-          : []
+      afterResponse: [
+        (response) => {
+          if (response.body.extensions && response.body.extensions.cost) {
+            this.updateGraphqlLimits(response.body.extensions.cost);
+          }
+
+          if (this.options.maxRetries === 0) {
+            return response;
+          }
+
+          if (response.body && response.body.errors) {
+            const retryDelay = maybeRetryMS({ response });
+
+            if (retryDelay) {
+              const first = response.body.errors[0];
+              const err = new Error(first.message);
+
+              err.locations = first.locations;
+              err.path = first.path;
+              err.extensions = first.extensions;
+              err.response = response;
+
+              // Make got consider this response errored and retry
+              throw err;
+            }
+          }
+
+          return response;
+        }
+      ]
     }
   };
 
   return got(uri, options).then((res) => {
-    if (res.body.extensions && res.body.extensions.cost) {
-      this.updateGraphqlLimits(res.body.extensions.cost);
-    }
-
     if (res.body.errors) {
       const first = res.body.errors[0];
       const err = new Error(first.message);
@@ -450,29 +451,13 @@ function maybeRetryMS(error) {
 }
 
 /**
- * Return a duration after which to retry a graphql call, or null if it
- * shouldn't be retried
- *
- * @param {Number} attemptNumber Shopify graphql call attempt no
- * @param {Number} maxRetries Max retries in default option
- * @param {Object} error Error object from got call
- * @return {Boolean | undefined}
- * @private
- */
-function graphQLRetryMS(attemptNumber, maxRetries, error) {
-  if (attemptNumber <= maxRetries) {
-    return maybeRetryMS(error);
-  }
-}
-
-/**
- * Got `calculateDelay` hook function passed when calling the REST API
+ * Got `calculateDelay` hook function.
  *
- * @param {Object} retryObject got's input for the retry logic
+ * @param {Object} retryObject Got's input for the retry logic
  * @return {Number}
  * @private
  */
-function calculateRESTDelay(retryObject) {
+function calculateDelay(retryObject) {
   return maybeRetryMS(retryObject.error) || retryObject.computedValue;
 }
```

Also note the calls to `shopify.updateLimits()` and `shopify.updateGraphqlLimits()` inside the `afterResponse` hook to address https://github.com/MONEI/Shopify-api-node/issues/251#issuecomment-487965284.

lpinca (approved):

lpinca:
Thank you.

PR #542, opened 2022-06-23
chore: use a more revealing assertion for leftover nock mocks

airhorns:
There's a wise afterEach assertion that runs after every test to ensure that no nock request mocks are left over. If you're working on tests which use nock, it can be annoying though, because you need to figure out which mock is not being hit to fix the test. Instead of just asserting that there are no mocks, we can use an assertion of the list of mocks to show the developer what mocks are still hanging around with mocha's nice diff.


lpinca (approved):

lpinca:
Thank you.

PR #554, opened 2022-09-19
Fix retryable status codes not respecting max retries option

jasong689:
While writing some tests for our own usage of `shopif-api-node`, I noticed that retries on `retryableStatusCodes` were not respecting `got`'s retry limit. According to `got`'s documentation:

> This function is responsible for the entire retry mechanism, including the limit property. To support this, you need to check if computedValue is different than 0.

Let's rely on their recommended check to ensure we aren't past the retry limit and ensure that retries are not indefinite if a `maxRetries` is set. Lastly, since we haven't explicitly included status code `200` as retryable we have to handle "Throttled" requests before attempting to check `computedValue`.

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,10 +19,11 @@ const retryableErrorCodes = new Set([
   'ENETUNREACH',
   'EAI_AGAIN'
 ]);
+const retryableErrorCodesArray = Array.from(retryableErrorCodes);
 
-const retryableStatusCodes = new Set([
+const retryableStatusCodesArray = [
   408, 413, 429, 500, 502, 503, 504, 521, 522, 524
-]);
+];
 
 /**
  * Creates a Shopify instance.
@@ -159,7 +160,10 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify `Retry-After` header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay
+            calculateDelay,
+            methods: [method],
+            statusCodes: retryableStatusCodesArray,
+            errorCodes: retryableErrorCodesArray
           }
         : 0,
     method,
@@ -275,7 +279,10 @@ Shopify.prototype.graphql = function graphql(data, variables) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify `Retry-After` header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay
+            calculateDelay,
+            methods: ['POST'],
+            statusCodes: retryableStatusCodesArray,
+            errorCodes: retryableErrorCodesArray
           }
         : 0,
     hooks: {
@@ -303,18 +310,6 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 };
 
 resources.registerAll(Shopify);
-
-/**
- * Got `calculateDelay` hook function passed to decide how long to wait before
- * retrying.
- *
- * @param {Object} retryObject Got's input for the retry logic
- * @return {Number} The delay
- * @private
- */
-function calculateDelay(retryObject) {
-  return maybeRetryMS(retryObject.error) || retryObject.computedValue;
-}
 
 /**
  * Decorates an `Error` object with details from GraphQL errors in the response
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -393,6 +385,27 @@ function maybeRetryMS(error) {
     }
   }
 
+  // If the retry limit has been reached, abort the request
+  if (computedValue == 0 && retryOptions.limit > 0) {
```
lpinca, line 389:
```suggestion
  if (computedValue === 0) {
```

`retryOptions.limit` is guaranteed to be bigger than zero, otherwise `maybeRetryMS()` is not called.

test/shopify.test.js
```diff
@@ -573,6 +573,22 @@ describe('Shopify', () => {
       });
     });
 
+    it('retries 429 a maximum up to maxRetries', () => {
```
lpinca, line 576:
```suggestion
    it('honors the maxRetries option', () => {
```

index.js
```diff
@@ -351,31 +353,21 @@ function delay(ms) {
...
  * @private
  **/
-function maybeRetryMS(error) {
-  // For simplicity, retry network connectivity issues after a hardcoded 1s.
-  if (retryableErrorCodes.has(error.code)) {
-    return 1000;
-  }
-
+function maybeRetryMS(retryObject) {
+  const { error, computedValue, retryOptions } = retryObject;
```
lpinca, line 361:
```suggestion
  const { error, computedValue } = retryObject;
```

lpinca:
Can you please address the coverage issue?

jasong689:
Will take a look at it

index.js
```diff
@@ -159,7 +159,10 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify `Retry-After` header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay
+            calculateDelay,
+            methods: [method],
+            statusCodes: Array.from(retryableStatusCodes),
```
jasong689, line 164: Instead of having separate retryable codes/errors between `got` and `shopify-api-node`, we should pass our expected retryable codes and errors to `got`

index.js
```diff
@@ -351,31 +357,21 @@ function delay(ms) {
...
-
+function maybeRetryMS(retryObject) {
+  const { error, computedValue } = retryObject;
   const response = error.response;
 
-  if (response.headers && response.headers['retry-after']) {
-    return response.headers['retry-after'] * 1000 || null;
-  }
-
-  if (retryableStatusCodes.has(response.statusCode)) {
```
jasong689, line 370: This check shouldn't be needed anymore if we pass the same retryable status codes to `got` that we have defined in `shopify-api-node`

jasong689:
@lpinca could you run the workflow again to see if the coverage issue has been resolved

index.js
lpinca: I'm not a fan of this. Copying the array for every request does not seem a good idea. The same applies to the line below.

index.js
```diff
@@ -351,31 +357,21 @@ function delay(ms) {
...
-    // Arbitrary 2 seconds, in case we get a 429 without a `Retry-After`
-    // response header, or 4xx/5xx series error that matches the Got retry
-    // defaults.
-    return 2 * 1000;
```
lpinca, line 374: We might return `null` instead and use the default `computedValue`. Another option is to still return 2 sec and remove the unreachable code (the last `return null` and the fallback `retryObject.computedValue` in `calculateDelay()`).

index.js
jasong689: I will update this to be a `const` at the module level a re-use the same array for each call

index.js
```diff
@@ -159,7 +157,10 @@ Shopify.prototype.request = function request(uri, method, key, data, headers) {
             limit: this.options.maxRetries,
             // Don't clamp Shopify `Retry-After` header values too low.
             maxRetryAfter: Infinity,
-            calculateDelay
+            calculateDelay,
+            methods: [method],
+            statusCodes: retryableStatusCodes,
+            errorCodes: Array.from(retryableErrorCodes)
```
lpinca, line 163: Avoid copying please.

index.js
```diff
@@ -393,7 +387,24 @@ function maybeRetryMS(error) {
     }
   }
 
-  return null;
+  // If the retry limit has been reached, abort the request
```
lpinca, line 390:
```suggestion
  // Stop retrying if the attempt limit has been reached or the request is not retryable.
```

index.js
```diff
@@ -313,7 +320,7 @@ resources.registerAll(Shopify);
  * @private
  */
 function calculateDelay(retryObject) {
-  return maybeRetryMS(retryObject.error) || retryObject.computedValue;
+  return maybeRetryMS(retryObject);
```
jasong689, line 323: Do you want me to rename `maybeRetryMS` to `calculateDelay` and remove this?

index.js
```diff
@@ -19,10 +19,11 @@ const retryableErrorCodes = new Set([
   'ENETUNREACH',
   'EAI_AGAIN'
 ]);
+const retryableErrorCodesArray = Array.from(retryableErrorCodes);
```
lpinca, line 22: I would just convert `retryableErrorCodes` to an array instead of using `Array.from()`, and the same below for status codes.

jasong689: Error codes have a different delay (1s) so there's a `retryableErrorCodes.has()` check. We can do `retryableErrorCodes.includes()` if we switch over to an array?

index.js
lpinca: Yes, it makes sense.

index.js
```diff
@@ -351,31 +358,21 @@ function delay(ms) {
  * Given an error from Got, see if Shopify told us how long to wait before
  * retrying.
  *
- * @param {Object} error Error object from Got call
+ * @param {Object} retryObject Got's input for the retry logic
  * @return {(Number|null)} The duration in ms, or `null`
```
lpinca, line 362:
```suggestion
 * @return {Number} The duration in ms
```

lpinca (approved):

lpinca:
Thank you.

PR #560, opened 2022-12-03
Add support for passing request lifecycle hooks down to got

airhorns:
`got` has rich support for request lifecycle hooks that allow userland code to understand what is happening within got and to modify it's behaviour. `shopify-api-node` uses got hooks for some retry implementation, but I think that users of `shopify-api-node` like me also want to know about whats happening during retries and errors. 

For this reason, I think that `shopify-api-node` should accept request hooks too! This allows users of shopify-api-node to log each retry, or decorate errors with extra context, or add in tracing and other observability hooks to get great visibility on what the library is doing. See a good example of this in the readme.

There's not much maintenance burden for this library if we just passthrough to got (like we do with the request options itself).

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,11 @@ Creates a new `Shopify` instance.
   for requests to the REST API, and the throttled cost data for requests to the
   GraphQL API, and retry the request after that time has elapsed. Mutually
   exclusive with the `autoLimit` option.
+- `hooks` - Optional - A list of `got`
+  [request hooks](https://github.com/sindresorhus/got/tree/v11#hooks) to attach
+  to all outgoing requests, like `beforeRetry`, `afterResponse`, etc. Hooks
+  should be provided in the same format that Got expects them and will receive
+  the same arguments Got passes unchanged.
 
 #### Return value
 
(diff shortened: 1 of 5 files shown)


lpinca:
I'm not a fan of this. It makes the library too coupled with the got dependency. If users starts using this feature, it will be hard to change got for something else if we ever need to do it. Similarly, it will be hard to update got if the hook mechanism will change in future versions of got.

lpinca:
Let me think about it. High level API wrappers usually do not provide this feature.

> how likely do you think swapping out got really is

Unlikely but not impossible. I thought about using `fetch` several times in the past.

> and if you want to swap out got for a different library, is it really that likely to be a minor version that doesn't involve other interface changes?

We would need a major version because of got errors but that is a small breaking change compared to this.

> Also notable is that newer versions of got use ESM, so this library would have to update to use ESM as well in order to upgrade got major versions.

I know, this is the reason why got@11 is used.

lpinca:
Landed as https://github.com/MONEI/Shopify-api-node/commit/eeda19880b634d5080e3ab2cc5b77513276c8dbe.

airhorns:
Thanks @lpinca

PR #563, opened 2022-12-21
[api] Update fulfillment endpoints

tkalliom:
- Add fulfillment.cancelV2 and fulfillment.listV2
- Document old cancel, complete, create, open, and update as deprecated

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -385,6 +385,9 @@ declare class Shopify {
     setFulfillmentOrdersDeadline: (
       params: Shopify.ISetFulfillmentOrdersDeadline
     ) => Promise<void>;
+    fulfillments: (
+      id: number
+    ) => Promise<Shopify.IPaginatedResult<Shopify.IFulfillment>>;
   };
   fulfillmentRequest: {
     accept: (
--- a/index.d.ts
+++ b/index.d.ts
@@ -319,6 +319,7 @@ declare class Shopify {
   };
   fulfillment: {
     cancel: (orderId: number, id: number) => Promise<Shopify.IFulfillment>;
+    cancelV2: (id: number) => Promise<Shopify.IFulfillment>;
     complete: (orderId: number, id: number) => Promise<Shopify.IFulfillment>;
     count: (orderId: number, params?: any) => Promise<number>;
     create: (orderId: number, params: any) => Promise<Shopify.IFulfillment>;
--- a/resources/fulfillment.js
+++ b/resources/fulfillment.js
@@ -69,6 +69,20 @@ Fulfillment.prototype.cancel = function cancel(orderId, id) {
 };
 
 /**
+ * Cancels a fulfillment.
+ *
+ * @param {Number} id Fulfillment ID
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+Fulfillment.prototype.cancelV2 = function cancelV2(id) {
+  const url = base.buildUrl.call(this, `${id}/cancel`);
+  return this.shopify
+    .request(url, 'POST', undefined, {})
+    .then((body) => body[this.key]);
+};
+
+/**
  * Creates a fulfillment for one or many fulfillment orders. The fulfillment
  * orders are associated with the same order and are assigned to the same
  * location.
--- a/test/fulfillment.test.js
+++ b/test/fulfillment.test.js
@@ -142,7 +142,7 @@ describe('Shopify#fulfillment', () => {
       .then((data) => expect(data).to.deep.equal(output.fulfillment));
   });
 
-  it('cancels a pending fulfillment', () => {
+  it('cancels a fulfillment for a specific order ID', () => {
     const output = fixtures.res.cancel;
 
     scope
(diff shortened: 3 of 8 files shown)


resources/fulfillment.js
```diff
@@ -82,6 +99,23 @@ Fulfillment.prototype.createV2 = function createV2(params) {
...
 };
 
+/**
+ * Lists fulfillments for a fulfillment order.
+ *
+ * @param {Number} fulfillmentOrderId Fulfillment order ID
+ * @return {Promise} Promise that resolves with the result
+ * @public
+ */
+Fulfillment.prototype.listV2 = function listV2(fulfillmentOrderId) {
```
tkalliom, line 109: Alternatively, this could be `fulfillmentOrder.fulfillments` instead of `fulfillment.listV2` – which is better?

FlavioAandres:
Hey guys, I'm trying to migrate the use cases for the new 2023-01 API update, and I just noticed we need to use the following: 

- `POST /admin/api/2022-10/fulfillments.json` instead of 
- `POST /orders/{order_id}/fulfillments.json` 

Do you know if this library is supporting this change?? Thanks!!! 
Ref: https://shopify.dev/api/release-notes/2022-07

resources/fulfillment.js
lpinca: It should be on `FulfillmentOrder` but this https://shopify.dev/api/admin-rest/2022-10/resources/fulfillmentorder#get-orders-order-id-fulfillment-orders already exists as the current `fulfillmentOrder.list()` so I'm not sure if this is needed.

test/fulfillment.test.js
```diff
@@ -154,6 +166,18 @@ describe('Shopify#fulfillment', () => {
       .then((data) => expect(data).to.deep.equal(output.fulfillment));
   });
 
+  it('cancels a pending fulfillment of a FulfillmentOrder', () => {
```
lpinca, line 169:
```suggestion
  it('cancels a fulfillment', () => {
```

Also please change the title of the previous test to `'cancels a fulfillment for a specific order ID'`

test/fulfillment.test.js
```diff
@@ -154,6 +166,18 @@ describe('Shopify#fulfillment', () => {
...
+    const output = fixtures.res.cancel;
+
+    scope
+      .post('/admin/fulfillments/255858046/cancel.json', {})
+      .reply(201, output);
```
lpinca, line 174:
```suggestion
      .reply(200, output);
```

lpinca:
@FlavioAandres the `/orders/{order_id}/fulfillments.json` seems to be still supported. See https://shopify.dev/api/admin-rest/2023-01/resources/fulfillmentorder#get-orders-order-id-fulfillment-orders.

Can you test it?

```js
const shopify = new Shopify({
  shopName: 'xxx',
  apiVersion: '2023-01',
  accessToken: 'xxx'
});

const fulfillments = await shopify.fulfillment.list(orderId);
```

lpinca:
@FlavioAandres I misread your comment, sorry. You can use:

```js
const fulfillment = await shopify.fulfillment.createV2(params);
```

index.d.ts
```diff
@@ -318,9 +318,19 @@ declare class Shopify {
     list: (params?: any) => Promise<Shopify.IPaginatedResult<Shopify.IEvent>>;
   };
   fulfillment: {
+    /**
+     * @deprecated
+     */
```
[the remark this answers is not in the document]
tkalliom: It exists in the older versions, yes, but already in `2022-01` these were deprecated and the new endpoints existed; see https://shopify.dev/api/admin-rest/2022-01/resources/fulfillment#post-orders-order-id-fulfillments-fulfillment-id-cancel. IMHO since these are deprecated in all API versions currently supported, it would make sense to mark them deprecated in the library as well, but of course these tags can be dropped at maintainer’s discretion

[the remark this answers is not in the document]
lpinca: There are other endpoints beside these that are deprecated/removed. I think that the JSDoc `@deprecated` tag here is not very useful. The final user does not see it and it gives the developer the impression that the endpoint still exist and this is not always true. I think it is better to rely on Shopify's warnings and return values.

lpinca:
Can you please split the commit into two?

- `[api] Add cancelV2 action to Fulfillment resource`
- `[api] Add fulfillments actions to FulfillmentOrder`

lpinca (approved):

lpinca:
Thank you.

FlavioAandres:
@tkalliom Thanks for adding those changes, are really useful. 
@lpinca Do you have an idea when these changes can be released to NPM?

lpinca:
@FlavioAandres soon. I would like to also include https://github.com/MONEI/Shopify-api-node/pull/560 and https://github.com/MONEI/Shopify-api-node/pull/564 in the next release.

kristiandm:
Hello, I'm kinda new to using this library so I apologize if my question is a bit weird. Can someone explain to me how to use createV2? we've been using create for a while but because of the 2023-01 change were trying to make createV2 work. What is the difference in using create and createV2 regarding the params? and how does createV2 know which order is for fulfillment without the orderId parameter like the create function has?

PR #564, opened 2022-12-22
graphql: capture first errors message for 2xx error responses only

summersk:
Error improvement for GraphQL API error responses per docs: https://shopify.dev/api/admin-graphql#status_and_error_codes

**Issue**

HTTP error responses return `{ errors: '...' }` instead of `{ errors: [...] }`

This results in errors with `message: undefined` for some errors.

Example responses:
```
HTTP/1.1 402 Payment Required
{ "errors": "This shop's plan does not have access to this feature" }
```
```
HTTP/1.1 404 Not Found
{ "errors": "Not Found" }
```

**Change**

~Fallback to passing through `body.errors` for non array values. Avoid branching on http status codes as it's not entirely clear what segments need handling (eg. 3xx).~

Ensure non-2xx errors _with_ a `errors` property that isn't an array are not intercepted and pass through as response errors.

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,7 +295,7 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
             if (res.body.errors) {
               // Make Got consider this response errored and retry if needed.
-              throw new Error(res.body.errors[0].message);
+              throw new Error(Array.isArray(res.body.errors) ? res.body.errors[0].message : res.body.errors);
             }
           }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -295,7 +295,11 @@ Shopify.prototype.graphql = function graphql(data, variables) {
 
             if (res.body.errors) {
```
lpinca, line 296: I think it's better to add the `Array.isArray()` check here, in the `if` condition. This branch should only cover the 2xx case. The same check should also be done in the `if` condition on line 323.

summersk: Sure thing, updated to check the response body and applied same check to `decorateError`

lpinca (approved):

lpinca:
Thank you.

PR #568, opened 2023-01-03
Fix #566 and introduce light types tests

### LINKED ISSUE #566: 'Got' only refers to a type but used as namespace
Hello,

Installing the latest version of shopify-api-node, I now receive these errors when trying to build my project:

`'Got' only refers to a type, but is being used as namespace here`

image

airhorns:
This fixes a type error introduced in https://github.com/MONEI/Shopify-api-node/commit/eeda19880b634d5080e3ab2cc5b77513276c8dbe and reported in https://github.com/MONEI/Shopify-api-node/issues/566 . I made this mistake by copying code suitable for got 12 into a codebase using got 11 -- my apologies! To avoid breakages like this, I think it'd be good to have a lightweight typescript test suite that verifies the types file works as expected. tsd is a great tool for this so I added it here as well.

### DIFF
--- a/index.test-d.ts
+++ b/index.test-d.ts
@@ -1,0 +1,43 @@
+import { expectType } from 'tsd';
+import Shopify from '.';
+
+// can be constructed with public access token
+new Shopify({
+  shopName: 'my-shopify-store.myshopify.com',
+  accessToken: '111'
+});
+
+// can be constructed with public access token and version
+const client = new Shopify({
+  shopName: 'my-shopify-store.myshopify.com',
+  accessToken: '111',
+  apiVersion: '2020-01'
+});
+
+// can be constructed with got hooks
+new Shopify({
+  shopName: 'my-shopify-store.myshopify.com',
+  accessToken: '111',
+  hooks: {
+    beforeRequest: [
+      (options) => {
+        options.headers['X-Test'] = 'test';
+      }
+    ]
+  }
+});
+
+expectType<number>(client.callLimits.remaining);
+expectType<number>(client.callLimits.current);
+expectType<number>(client.callLimits.max);
+
+const shop = await client.shop.get();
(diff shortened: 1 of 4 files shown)


index.test-d.ts
```diff
@@ -0,0 +1,43 @@
+import { expectType } from 'tsd';
```
lpinca, line 1: Can this be moved to the `test` folder?

index.test-d.ts
```diff
@@ -0,0 +1,43 @@
...
+import Shopify from '.';
+
+// can be constructed with public access token
```
lpinca, line 4:
```suggestion
// Can be constructed with public access token.
```

index.test-d.ts
```diff
@@ -0,0 +1,43 @@
...
+new Shopify({
+  shopName: 'my-shopify-store.myshopify.com',
+  accessToken: '111'
+});
+
+// can be constructed with public access token and version
```
lpinca, line 10:
```suggestion
// Can be constructed with public access token and version.
```

index.test-d.ts
```diff
@@ -0,0 +1,43 @@
...
+const client = new Shopify({
+  shopName: 'my-shopify-store.myshopify.com',
+  accessToken: '111',
+  apiVersion: '2020-01'
+});
+
+// can be constructed with got hooks
```
lpinca, line 17:
```suggestion
// Can be constructed with got hooks.
```

index.test-d.ts
lpinca: I wonder if it is better to move both `index.d.ts` and `index.test-d.ts` in a new `types` folder.

lpinca:
Landed in cfafdf1ce006c5e9...a329b5ecf890be01.

lpinca:
Thank you.

PR #574, opened 2023-01-19
delivery_method can be null

ikudosi:
Docs does not say it, but after playing around with the fulfillment orders API, there were some instances where it came back as null.


lpinca (approved):

lpinca:
Thank you.

Issue #585, opened 2023-03-06
[Improvement]: implementation of delay function without setTimeout


jeevan-jp:
Since setTimeout doesn't guarantee the delay of specified ms, the implementation of delay function in index.js can be improved with the help of while loop.
```js
function delay(ms) {
  return new Promise((resolve) => {
    const now = new Date().getTime();
    while (new Date().getTime() < now + ms) {
      /* Do nothing */
    }
    resolve(null); 
  });
}
```

Reference Reading: https://www.gimtec.io/articles/settimeout-does-not-guarantee-time-to-execution/

lpinca:
I'm closing this as per https://github.com/MONEI/Shopify-api-node/pull/586#issuecomment-1458841811.

PR #586, opened 2023-03-06
[Improvement]: implementation of delay function without setTimeout

jeevan-jp:
while loop has been used to improve delay function

### DIFF
--- a/index.js
+++ b/index.js
@@ -370,7 +370,13 @@ function decorateError(error) {
  * @private
  */
 function delay(ms) {
-  return new Promise((resolve) => setTimeout(resolve, ms));
+  return new Promise((resolve) => {
+    const now = new Date().getTime();
+    while (new Date().getTime() < now + ms) {
+      /* Do nothing */
+    }
+    resolve(null);
+  });
 }
 
 /**


jeevan-jp:
fixes #585

lpinca:
This is very wrong as it blocks the event loop. Run this

```
const promise = new Promise(function (resolve) {
  const now = new Date().getTime();
  while (new Date().getTime() < now + 5000) {
    /* Do nothing */
  }
  resolve(null);
});

console.log('OK');
```

and note how "OK" is printed _after_ 5 seconds. It should be printed immediately.

jeevan-jp:
Yeah, just went through this repo's delay function calls. You are right, blocking main thread would affect further processing by caller function.

I am still wondering whether there is a better implementation, since accuracy is something that's not guaranteed with setTimeout.

Seems there is none.

lpinca:
> I am still wondering whether there is a better implementation, since accuracy is something that's not guaranteed with setTimeout.

It is a very small inaccuracy (few milliseconds at most) so I think it does not matter.

lpinca:
I'm closing this. Thank you anyway.

PR #595, opened 2023-03-30
[ts] Add admin_graphql_api_id to ts types

peterz0:
Add admin_graphql_api_id to IOrder, ICustomer and IProduct.
This field is documented in the response example of the doc [1][2][3]

Ref:
1. https://shopify.dev/docs/api/admin-rest/2023-04/resources/customer
2. https://shopify.dev/docs/api/admin-rest/2023-04/resources/order
3. https://shopify.dev/docs/api/admin-rest/2023-04/resources/product

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1387,6 +1387,7 @@ declare namespace Shopify {
   interface ICustomer {
     accepts_marketing: boolean;
     addresses?: ICustomerAddress[];
+    admin_graphql_api_id: string;
     created_at: string;
     currency: string;
     default_address: ICustomerAddress;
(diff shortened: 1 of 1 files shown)


lpinca:
There are many more resources that have the `admin_graphql_api_id` property in the HTTP response data. I would prefer to not add it because:

1. It is not officially documented
2. The interface is usually used for both the HTTP request and response body.

peterz0:
The problem we want to solve is that we have a customer json of Shopify.ICustomer type (e.g. from customers webhook). Then we want to use admin_graphql_api_id for graphql API call.
However, we face type error because admin_graphql_api_id isn't in the ICustomer interface.
Do you have any advise on it? Thanks!

lpinca:
Not really. How about making the property optional and adding it to all resources where it exists?

peterz0:
Thanks! Let me close the pr at the moment.

PR #599, opened 2023-04-01
updated apis related to metafields

utkarsh-24:


lpinca:
Thanks but the current API works and this does not seem better.

utkarsh-24:
Please tell what is not appropriate .
When I forked this repo at that time metafield.js was not having much code.

Issue #606, opened 2023-06-05
Cannot get products metafields


aurelien-robineau:
Hi! I need to retrieve metafields for a product, but it returns the shop metafields.
My code:
```
const productMetafields = await shopify.metafield.list({
    owner_resource: 'product',
    owner_id: lineItem.product_id,
    namespace: 'global',
    limit: 250
});
```

All the returned metafields have the `owner_resource` set to `shop` and are indeed shop metafields.

Package version: `3.12.5`.

Any clue on why this is happening ?

Thank you!

lpinca:
See https://github.com/MONEI/Shopify-api-node#metafields. You are missing the `metafield` key in your option object.

lpinca:
I'm closing this as answered.

PR #607, opened 2023-06-05
Fix issue with metafields

aurelien-robineau:
Metafields are child resources, so must use the `base-child` mixin.

For example, the metafields list function only returns the Shop metafields, regardless of the `owner_resource` and `owner_id` that we provide.

### DIFF
--- a/resources/metafield.js
+++ b/resources/metafield.js
@@ -2,7 +2,7 @@
 
 const assign = require('lodash/assign');
 
-const base = require('../mixins/base');
+const baseChild = require('../mixins/base-child');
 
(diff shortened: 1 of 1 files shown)


lpinca:
This does not work. If we do this we have to create a metafield resource for each parent resource. `ArticleMetafield`, `BlogMetafield`, etc. `Metafield` used to be a top-level resource and AFAIK it still works as top-level resource.

lpinca:
I'm closing this as per https://github.com/MONEI/Shopify-api-node/issues/606.

PR #628, opened 2023-08-23
Proposing ApiVersion enum

TheRealFlyingCoder:
It seems unusual to have the typescript definitions but leave API Version as string given there's a fixed set of options.

Unsure about your naming conventions, but this PR is just what i'd expect from a package:

1. Export enum of available ApiVersion's
2. Take string | ApiVersion enum, so you can still build if the enum becomes outdated with later releases.

```
import Shopify, { ApiVersion } from 'shopify-api-node';

const client = new Shopify({
    accessToken,
    shopName,
    apiVersion: ApiVersion.July23
})
```

Happy to make any tweaks

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -780,9 +780,20 @@ declare namespace Shopify {
     interval: number;
   }
 
+  export declare enum ApiVersion {
+    April22 = "2022-04",
+    July22 = "2022-07",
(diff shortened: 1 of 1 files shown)


lpinca:
I don't find it very useful and keeping it is sync would be a maintenance burden.

Issue #506, opened 2021-08-26
Use a logical approach for resource modules (ie: stop using dynamic imports)


panoply:
Why do this?
https://github.com/MONEI/Shopify-api-node/blob/master/resources/index.js

The implications for loading modules in this manner for exceed the benefits. Bundlers are unable to efficiently treeshake and those mapped dynamic requires wreak absolute havoc when one attempts to re-route them in transpilation. Please, move to ESM or are the very least process the project through a bundler that prevents that dynamic mapping.

reecefenwick:
Facing this issue as well, are there any implications with changing from lazy/dynamic imports?

I understand the rationale, but feel it could be an unnecessary optimisation. I'm working in large server side project with a large dependency tree and this is the only dependency I have that is not compatible for bundling

reecefenwick:
I've raised https://github.com/MONEI/Shopify-api-node/pull/634 to fix this with backwards compatibility

PR #634, opened 2023-10-09
fix/bundler-friendly-lazy-imports

reecefenwick:
Changes the dynamic imports to be bundler friendly, but still remain as lazy imports.

Fixes https://github.com/MONEI/Shopify-api-node/issues/506

Test suite passes locally ✅ 
Tested this with esbuild that was previously failing ✅ 

Also addresses other issues out in the wild
* https://github.com/aws/aws-sam-cli/issues/4550


resources/index.js
```diff
@@ -80,18 +80,18 @@ const map = {
...
  */
 function registerAll(Shopify) {
-  Object.keys(map).forEach((prop) => {
-    Object.defineProperty(Shopify.prototype, prop, {
+  Object.entries(map).forEach(([resourceName, importResource]) => {
+    Object.defineProperty(Shopify.prototype, resourceName, {
       configurable: true,
       get: function get() {
-        const resource = require(`./${map[prop]}`);
+        const resource = importResource();
```
lpinca, line 87: I don't understand the difference. What bundler issue is this fixing? And why a wrapper around `require()` fixes it? Also, can something like this be used

```js
function importResource(name) {
  return require(name);
}
```

instead of defining a dedicated function for each resource?

reecefenwick: Bundlers work by traversing code, resolving imports/requires until there are none left, adding them to a bundle as it goes

The problem we are facing when bundling is due to the use of a dynamic token when importing`${map[prop]}`. Bundlers aren't smart enough to know the context and how to resolve that import, especially when in a loop

The fix is to explicitly define the lazy imports (or require) in a way that can be analysed statically by bundlers. Which means we need to define imports like so `require('./payout')` so that bundlers can successfully resolve them

While it's more verbose the outcome is exactly the same, lazy imports at runtime. But with the benefit of being friendly for bundling :)

lpinca: > Bundlers aren't smart enough to know the context and how to resolve that import, especially when in a loop

What bundler specifically? I think it works with some of them.

reecefenwick: You're right, it does work with webpack IIRC.

I am using esbuild, which doesn't support this pattern of dynamic imports

Do you have any concerns with this change? It is 100% backwards compatible with identical behaviour, but solves a problem a few people have encountered

lpinca (approved):

lpinca:
Thank you.

PR #654, opened 2024-06-26
Fixed type of ICustomer.multipass_identifier

natewaddoups:
I suspect that "multipass_identifier: null" was never the intent...


natewaddoups:
There's a bigger problem here - the createCustomer mutation expects an object with camelCased properties, but the shopify-api-node library requires an object with snake_cased properties. 

I'll send another PR to fix both issues.

lpinca (approved):

lpinca:
Thank you.

Issue #669, opened 2024-12-20
Unsupported Content-Type header, use application/json 2025-01


mete89:
I have a simple graphql which returns `Unsupported Content-Type header, use application/json 2025-01` error in 2025-01 although it works fine versions <= 2024-10. 

```
            {
                productVariantsCount {
                  count
                  precision
                }
            }
```

It works fine when we add an empty object as variable to the graphql to force json usage here:
```
Shopify.prototype.graphql = function graphql(data, variables) {
  let pathname = '/admin/api';

  if (this.options.apiVersion) {
    pathname += `/${this.options.apiVersion}`;
  }

  pathname += '/graphql.json';

  const uri = { pathname, ...this.baseUrl };
  const json = variables !== undefined && variables !== null;
  const options = {
    agent: this.options.agent,
    body: json ? this.options.stringifyJson({ query: data, variables }) : data,
    headers: {
      ...this.baseHeaders,
      'Content-Type': json ? 'application/json' : 'application/graphql'
    },
    method: 'POST',
    parseJson: this.options.parseJson,
    responseType: 'json',
    timeout: this.options.timeout
  };
```

Related changelog:

https://shopify.dev/docs/api/release-notes/2025-01

JSON body support: We now provide native support of a HTTP request body as a JSON_body for function external calls. Use this field when the body is in JSON format to reduce function instruction consumption and to ensure the body is formatted in logs. Don't use the JSON_body field together with the body field in the [HttpResponse](https://shopify.dev/docs/api/functions/reference/cart-checkout-validation/graphql/common-objects/httpresponse) object. If both are provided, then the HttpResponse's body field takes precedence. If the JSON_body field is specified and no Content-Type header is included, then the header will automatically be set to application/json.

RobWiddick:
+1 after trying to use 2025-01.

Bek81:
+1!

aldoalprak:
+1 just tried to use 2025-01 and got the same error

fagnanigm:
+1 - Yes, same problem here using 2025-01

infiton:
Happy friday everyone 😞 

https://community.shopify.dev/t/version-2025-01-has-broken-previously-valid-requests/7769

infiton:
ok Shopify did announce this change: https://shopify.dev/changelog/graphql-over-http

PR #670, opened 2025-02-07
resolves issue #669

infiton:
This issue resolves https://github.com/MONEI/Shopify-api-node/issues/669

Based on https://shopify.dev/changelog/graphql-over-http:

```
For requests to 2025-01 versions and after:

Use the /graphql endpoint. Using /graphql.json will automatically respond with application/json as it does now.
The request body must contain a [spec JSON encoding](https://graphql.github.io/graphql-over-http/draft/#sec-JSON-Encoding).
The Content-Type header must specify application/json. The legacy application/graphql request format is no longer supported. Requests with no supported content type will receive a 415 Unsupported Media Type response.
The Accept header now supports application/graphql-response+json and/or application/json response formats. Requests with no supported accept types will receive a 406 Not Acceptable response.
```

This change is backwards compatible as old api versions accept json requests like this

### DIFF
--- a/index.js
+++ b/index.js
@@ -273,13 +273,12 @@ Shopify.prototype.graphql = function graphql(data, variables) {
   pathname += '/graphql.json';
 
   const uri = { pathname, ...this.baseUrl };
-  const json = variables !== undefined && variables !== null;
   const options = {
     agent: this.options.agent,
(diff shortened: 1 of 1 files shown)


infiton:
@lpinca any thoughts on this?

lpinca (approved):

lpinca:
Thank you.

PR #673, opened 2025-04-05
allow for afterResponse hooks to run before graphql errors are thrown

infiton:
This PR is to implement @lpinca suggestion from https://github.com/MONEI/Shopify-api-node/pull/672 see that PR for the context on why this is desired


lpinca (approved):

lpinca:
Thank you.

